PR #13, opened 2017-07-18
Feature/mxapi

CaringLabs:
Add support for endings. Reading was working properly. Writing now works.


CaringLabs:
This should not be for master. Closing.

PR #19, opened 2019-04-07
Added missing mutex include in Decimals.cpp

RobinKa:
On Visual Studio 2017 I had the following error without this include

`MusicXML-Class-Library\Sourcecode\mx\core\Decimals.cpp(126): error C2039: 'mutex': is not a member of 'std'`


webern:
Thank you.

PR #53, opened 2020-02-17
cmake : Add install target. Cleanup.

p-groarke:
- Add install targets for both mx and ezxml.
- Ignore cmake generated files.
- Minor cleanup changes for my cmake OCD :)

- [x] Ran unit tests
- [x] Ran core tests

I had 1 failure in the tests, I'm unsure if it is expected or related to Windows. I could investigate if you want.
```
error in OrnamentsFunctions.cpp (line 155) mx::impl::OrnamentsFunctions::parseOrnament: 'This tremolo is not supported because it starts and stops on different notes.
```

edit : Only tested on Windows.

**Why are install targets useful/why do you want them?**
With the new install target, a user can run `cmake .. && cmake --build . --target install` which will install your library on the system. For example in `/usr/lib/...` on *nixes. It also install the necessary cmake files. So now users can consume mx with `find_package`.

For ex.
```cmake
find_package(Mx)
target_link_libraries(blee Mx)
```

It also makes mx work with `ExternalProject_Add`, which is what I use. `ExternalProject_Add` is a cmake function that clones a git repo, builds it and lets you link with it. So you don't have to fuss around with install paths or globally installing your dependencies.

**On an unrelated note :**
You may skip my following comments and roll your eyes ;) Buuuuut, I wouldn't personally add all the IDE files to your git repo. Since that's what cmake is for. By removing all that stuff, you could also have a much simpler `.gitignore`.

Embedding pugixml in your repo like you did will make mx unusable to anyone who links directly with pugixml in their project (duplicate symbols). You can google "dependency hell" for more info (yes, that's the real name). To fix that, you could use something super robust like conan to install pugi when you build mx. Or you can use the less robust `ExternalProject_Add` as well.  I would be happy to send a PR with any of those solutions.

Cheers


webern:
> I had 1 failure in the tests, I'm unsure if it is expected or related to Windows. I could investigate if you want.

I should remove that, it's not a failure, just a useless std::cout that is telling us that there are ornament constructs that are unsupported.

> Ran core tests

Wow, thank you, I hope you have a good CPU.

> I wouldn't personally add all the IDE files to your git repo

Not rolling my eyes, I know it's unpopular. At work, of course, I can't add `.idea` to a git repo, but for a 'just-for-fun project' where I'm the only maintainer, it's nice to have it there. Especially when there are things like auto code styling, run targets etc. in the `.idea` folder. Also the `Xcode` stuff actually has to be there because I use it to build iOS and macOS targets for [Komp](https://apps.apple.com/us/app/komp-create/id1103355632). AFIK these types of targets cannot (at least easily) be specified in Cmake.

Overall, I'm aware that it's not super-professional to drop a bunch of IDE turds, if I ever get into conflict with other maintainers, I'll remove that stuff. ;) For now, the first thing I do when I start a new project is bounce over to https:://gitignore.io and add some monstrosity like [this](https://www.gitignore.io/api/node,xcode,linux,macos,windows,webstorm,jetbrains). It's kind of a JavaScript-y thing to do.

> Embedding pugixml in your repo like you did will make mx unusable to anyone who links directly with pugixml in their project

Good point. I sort of hate `find_package`, `ExternalProject_Add` and such, I always want to build from vendored source.  I could just patch pugixml's namespace to avoid the issue. I realize with real projects you can't get away with it, but it's just a teeny tiny dependency here. With things like zlib, etc, you can't get away with it.

That being said:
> Add install targets for both mx and ezxml

Yes, just because I would vendor it doesn't mean others should have to, so, good idea. And cleanups are always good as well. I'll take a closer look soon.

Thanks.
Matt

Issue #70, opened 2020-03-28
[feature] mx::api NonTraditionalKey equivalent


p-groarke:
We've discussed about this in the past, and I've found mixing `mx::core` with `mx::api` to be a pain! :)

So, I'm just adding this here in case someday you'd want to add this feature to `mx::api`.

Cheers

webern:
If you'd rather I finish up with my branch (I think it's ready) then I can do that instead.

p-groarke:
Up to you. If you already have the feature up and working + tested, you can just merge that and close my PR. I won't be offended by it ;)

PR #77, opened 2020-03-28
Add NonTraditional support to KeyData.

p-groarke:
Fixes https://github.com/webern/mx/issues/70

As far as I could tell, this should be enough to support non-traditional keys in `api`.

However, I couldn't for the life of me figure out your unit test framework :) Could you explain to me how to add some tests for this? ty

Of note
- Using a vector of `PitchData`. Maybe you'd rather I add a new "Data" type which only contains step, alter and accidental? `NonTraditionalData` maybe?
- Minor tweaks to cmake.
- No tests yet! PR will stay Draft until I can add some good tests.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,7 +38,7 @@ file(GLOB_RECURSE SRC_MX_TEST_UTILITY ${PRIVATE_DIR}/mxtest/utility/*.cpp ${PRIV
 file(GLOB_RECURSE SRC_CPUL ${PRIVATE_DIR}/cpul/*.cpp ${SOURCE}/cpul/*.h)
 
 # Mx Library
-add_library(${PROJECT_NAME} STATIC ${SRC_MX_API} ${SRC_MX_CORE} ${SRC_MX_IMPL} ${SRC_MX_PUGIXML} ${SRC_MX_UTILITY})
+add_library(${PROJECT_NAME} STATIC ${HEADERS_MX_API} ${SRC_MX_API} ${SRC_MX_CORE} ${SRC_MX_IMPL} ${SRC_MX_PUGIXML} ${SRC_MX_UTILITY})
 source_group( "api-public" FILES ${HEADERS_MX_API})
 source_group( "api" FILES ${SRC_MX_API} )
 source_group( "core" FILES ${SRC_MX_CORE} )
@@ -103,6 +103,7 @@ if(MX_BUILD_TESTS)
     target_link_libraries(MxTest ${PROJECT_NAME})
     target_link_libraries(MxTest ${CMAKE_THREAD_LIBS_INIT})
     set_property(TARGET MxTest PROPERTY CXX_STANDARD 14)
+    set_property(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY VS_STARTUP_PROJECT MxTest)
 else()
     message(STATUS "${PROJECT_NAME}:${CMAKE_CURRENT_LIST_LINE} tests will NOT be compiled")
 endif()
--- a/Sourcecode/include/mx/api/KeyData.h
+++ b/Sourcecode/include/mx/api/KeyData.h
@@ -4,6 +4,10 @@
 
 #pragma once
 
+#include "mx/api/PitchData.h"
+
+#include <vector>
+
 namespace mx
 {
 	namespace api
@@ -49,6 +53,14 @@ namespace mx
             // within the part
             int staffIndex;
             
+            // If this is true, the KeyData is in "Non Traditional" mode.
+            // The the 'fifths' and 'mode' member variables will be ignored, and
+            // key data will be read from nonTraditionalPitchData.
+            bool nonTraditionalKey;
+            // The non-tradional accidentals.
+            // Fill this with your custom accidentals and alters.
+            std::vector<PitchData> nonTraditionalPitchData;
+
 			// TODO support position data and/or other attribtues
             
             KeyData()
@@ -57,6 +69,8 @@ namespace mx
             , mode{ KeyMode::unspecified }
             , tickTimePosition{ 0 }
             , staffIndex{ -1 }
+            , nonTraditionalKey{ false }
+            , nonTraditionalPitchData{}
             {
                 
             }
@@ -68,6 +82,8 @@ namespace mx
         MXAPI_EQUALS_MEMBER( mode )
         MXAPI_EQUALS_MEMBER( tickTimePosition )
         MXAPI_EQUALS_MEMBER( staffIndex )
+        MXAPI_EQUALS_MEMBER( nonTraditionalKey )
+        MXAPI_EQUALS_MEMBER( nonTraditionalPitchData )
         MXAPI_EQUALS_END;
         MXAPI_NOT_EQUALS_AND_VECTORS( KeyData );
 	}
--- a/Sourcecode/include/mx/api/PitchData.h
+++ b/Sourcecode/include/mx/api/PitchData.h
@@ -10,15 +10,16 @@ namespace mx
 {
     namespace api
     {
-    	enum class Step
-    	{
-    		c = 0,
-    		d = 1,
-    		e = 2,
-    		f = 3,
-    		g = 4,
-    		a = 5,
-    		b = 6,
+        enum class Step
+        {
+            c = 0,
+            d = 1,
+            e = 2,
+            f = 3,
+            g = 4,
+            a = 5,
+            b = 6,
+            count,
             unspecified = -1
     	};
 
--- a/Sourcecode/private/mx/impl/MeasureReader.cpp
+++ b/Sourcecode/private/mx/impl/MeasureReader.cpp
@@ -35,6 +35,7 @@
 #include "mx/core/elements/Grouping.h"
 #include "mx/core/elements/Harmony.h"
 #include "mx/core/elements/Key.h"
+#include "mx/core/elements/NonTraditionalKey.h"
 #include "mx/core/elements/KeyAccidental.h"
 #include "mx/core/elements/KeyAlter.h"
 #include "mx/core/elements/KeyChoice.h"
@@ -452,15 +453,8 @@ namespace mx
                 const auto keyChoiceObj = *key.getKeyChoice();
                 const auto keyType = keyChoiceObj.getChoice();
 
-                if( keyType == core::KeyChoice::Choice::nonTraditionalKey )
-                {
-                    // TODO - support non-traditional keys
-                    continue;
-                }
-                
                 api::KeyData keyData;
                 // TODO - do position attributes
-                const auto& traditionalKey = *keyChoiceObj.getTraditionalKey();
                 
                 if( attr.hasNumber )
                 {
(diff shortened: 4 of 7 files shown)


CMakeLists.txt
```diff
@@ -38,7 +38,7 @@ file(GLOB_RECURSE SRC_MX_TEST_UTILITY ${PRIVATE_DIR}/mxtest/utility/*.cpp ${PRIV
 file(GLOB_RECURSE SRC_CPUL ${PRIVATE_DIR}/cpul/*.cpp ${SOURCE}/cpul/*.h)
 
 # Mx Library
-add_library(${PROJECT_NAME} STATIC ${SRC_MX_API} ${SRC_MX_CORE} ${SRC_MX_IMPL} ${SRC_MX_PUGIXML} ${SRC_MX_UTILITY})
+add_library(${PROJECT_NAME} STATIC ${HEADERS_MX_API} ${SRC_MX_API} ${SRC_MX_CORE} ${SRC_MX_IMPL} ${SRC_MX_PUGIXML} ${SRC_MX_UTILITY})
```
p-groarke: So the headers show up in IDE. Makes it much easier to navigate for new devs.

CMakeLists.txt
```diff
@@ -103,6 +103,7 @@ if(MX_BUILD_TESTS)
     target_link_libraries(MxTest ${PROJECT_NAME})
     target_link_libraries(MxTest ${CMAKE_THREAD_LIBS_INIT})
     set_property(TARGET MxTest PROPERTY CXX_STANDARD 14)
+    set_property(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY VS_STARTUP_PROJECT MxTest)
```
p-groarke: Makes MxTest the default target in MSVC.

Sourcecode/include/mx/api/KeyData.h
```diff
@@ -49,6 +53,14 @@ namespace mx
             // within the part
             int staffIndex;
             
+            // If this is true, the KeyData is in "Non Traditional" mode.
+            // The the 'fifths' and 'mode' member variables will be ignored, and
+            // key data will be read from nonTraditionalPitchData.
+            bool nonTraditionalKey;
+            // The non-tradional accidentals.
+            // Fill this with your custom accidentals and alters.
+            std::vector<PitchData> nonTraditionalPitchData;
```
p-groarke: Maybe create a new "Data" struct. "NonTraditionalData"?

Sourcecode/include/mx/api/PitchData.h
```diff
@@ -10,15 +10,16 @@ namespace mx
...
+        enum class Step
+        {
+            c = 0,
+            d = 1,
+            e = 2,
+            f = 3,
+            g = 4,
+            a = 5,
+            b = 6,
+            count,
```
p-groarke: :)

Sourcecode/private/mx/impl/PropertiesWriter.cpp
```diff
@@ -109,6 +113,7 @@ namespace mx
         PropertiesWriter::PropertiesWriter( const core::PartwiseMeasurePtr& inPartwiseMeasure )
         : myProperties{ nullptr }
         , myPartwiseMeasure{ inPartwiseMeasure }
+        , myConverter{}
```
p-groarke: Added converter because its nice.

Sourcecode/private/mx/impl/PropertiesWriter.cpp
```diff
@@ -165,26 +169,59 @@ namespace mx
...
+
+            } else {
+                key->getKeyChoice()->setChoice(core::KeyChoice::Choice::nonTraditionalKey);
+
+                for (const api::PitchData& p : inKeyData.nonTraditionalPitchData)
+                {
+                    core::NonTraditionalKeyPtr k = core::makeNonTraditionalKey();
+                    if (p.step == api::Step::count || p.step == api::Step::unspecified)
+                    {
+                        // Would you prefer throw? assert?
```
p-groarke: Not sure what is your standard in terms of "bad things reporting". Continue, throw, assert, format c:/ ?

Sourcecode/private/mx/impl/PropertiesWriter.cpp
```diff
@@ -165,26 +169,59 @@ namespace mx
...
+
+                    if (p.accidental != api::Accidental::none)
+                    {
+                        k->setHasKeyAccidental(true);
+                        auto acc = std::make_shared<core::KeyAccidental>( myConverter.convert(p.accidental) );
+                        k->setKeyAccidental(acc);
+                    }
+
+                    if (p.alter != 0) {
+                        k->setKeyAlter(std::make_shared<core::KeyAlter>(core::Semitones{ double(p.alter) }));
```
p-groarke: Drops support for weird semitones :/ Maybe should use doubles in "NonTraditionalData"?

Sourcecode/private/mx/impl/PropertiesWriter.cpp
```diff
@@ -165,26 +169,59 @@ namespace mx
...
+                    key->getKeyChoice()->addNonTraditionalKey(k);
+                }
             }
+
             myProperties->addKey( key );
         }
         
         
-        void PropertiesWriter::writeTime( const api::TimeSignatureData& value )
+		void PropertiesWriter::writeTime(const api::TimeSignatureData& value)
```
p-groarke: oops

Sourcecode/private/mxtest/api/ApiMuAccidentals1ScoreData.h
```diff
@@ -91,8 +91,19 @@ namespace mxtest
         measure->timeSignature.beatType = 4;
         measure->timeSignature.isImplicit = true;
         measure->staves.emplace_back( StaffData{} );
+        measure->keys.emplace_back( KeyData{} );
```
p-groarke: OK, I tried adding a test. But this doesn't do anything. I found 2 test files which test keys, this and ApiLy43eScoreData. I don't get what this tests.

Sourcecode/include/mx/api/PitchData.h
webern: Fine, whatever. :)

CMakeLists.txt
webern: Do a rebase or merge from master once #79 lands and then reapply what you need to make the project work better in the IDE.

Sourcecode/include/mx/api/KeyData.h
webern: I had started, or maybe completed this here https://github.com/webern/mx/master/develop...feature/custom-key-signatures

As a reference to how I was thinking about it. Or you could check it out, kick the tires. The main difference is...

> Maybe create a new "Data" struct. "NonTraditionalData"?

Yes, I think that would be better than using PitchData.
I'm sorry that the structure of test files is hard to understand (I don't remember how/why it got that way). You can see in my branch that simply added a test file.

I think it's probably best to add a test file that sort of matches the name of the struct in the event that a better one doesn't seem to already exist.

A lot of the really funky names you see in the tests are related to specific test files.

Some xsd for reference:
```xml
	<xs:complexType name="key">
		<xs:annotation>
			<xs:documentation>The key type represents a key signature. Both traditional and non-traditional key signatures are supported. The optional number attribute refers to staff numbers. If absent, the key signature applies to all staves in the part. Key signatures appear at the start of each system unless the print-object attribute has been set to "no".</xs:documentation>
		</xs:annotation>
		<xs:sequence>
			<xs:choice>
				<xs:group ref="traditional-key"/>
				<xs:group ref="non-traditional-key" minOccurs="0" maxOccurs="unbounded"/>
			</xs:choice>
			<xs:element name="key-octave" type="key-octave" minOccurs="0" maxOccurs="unbounded">
				<xs:annotation>
					<xs:documentation>The optional list of key-octave elements is used to specify in which octave each element of the key signature appears.</xs:documentation>
				</xs:annotation>
			</xs:element>
		</xs:sequence>
		<xs:attribute name="number" type="staff-number"/>
		<xs:attributeGroup ref="print-style"/>
		<xs:attributeGroup ref="print-object"/>
	</xs:complexType>
```

```xml
	<xs:group name="non-traditional-key">
		<xs:annotation>
			<xs:documentation>The non-traditional-key group represents a single alteration within a non-traditional key signature. A sequence of these groups makes up a non-traditional key signature</xs:documentation>
		</xs:annotation>
		<xs:sequence>
			<xs:element name="key-step" type="step">
				<xs:annotation>
					<xs:documentation>Non-traditional key signatures can be represented using the Humdrum/Scot concept of a list of altered tones. The key-step element indicates the pitch step to be altered, represented using the same names as in the step element.</xs:documentation>
				</xs:annotation>
			</xs:element>
			<xs:element name="key-alter" type="semitones">
				<xs:annotation>
					<xs:documentation>Non-traditional key signatures can be represented using the Humdrum/Scot concept of a list of altered tones. The key-alter element represents the alteration for a given pitch step, represented with semitones in the same manner as the alter element.</xs:documentation>
				</xs:annotation>
			</xs:element>
			<xs:element name="key-accidental" type="accidental-value" minOccurs="0">
				<xs:annotation>
					<xs:documentation>Non-traditional key signatures can be represented using the Humdrum/Scot concept of a list of altered tones. The key-accidental element indicates the accidental to be displayed in the key signature, represented in the same manner as the accidental element. It is used for disambiguating microtonal accidentals.</xs:documentation>
				</xs:annotation>
			</xs:element>
		</xs:sequence>
	</xs:group>
```

Sourcecode/private/mx/impl/MeasureReader.cpp
```diff
@@ -470,33 +464,60 @@ namespace mx
...
+                    keyData.nonTraditionalKey = true;
+
+                    const auto& keyVec = keyChoiceObj.getNonTraditionalKeySet();
+
+                    for (const core::NonTraditionalKeyPtr& k : keyVec)
                     {
-                        keyData.mode = api::KeyMode::unsupported;
+                        api::PitchData p;
+                        p.step = myConverter.convert(k->getKeyStep()->getValue());
+                        p.alter = int(k->getKeyAlter()->getValue().getValue());
```
webern: Microtones will need to be handled in semitones (int) and cents (double).

Sourcecode/private/mx/impl/PropertiesWriter.cpp
webern: I think just silently fall back to `Step::c`. If I remember right I only use exceptions/assertions for bugs in the library, not for bad input.

Sourcecode/private/mxtest/api/ApiMuAccidentals1ScoreData.h
webern: Just add a new test for this `api/KeyDataTest.cpp`.

ApiLy43eScoreData is probably related to a test file by that name from the Lilypond MusicXML test suite.

Sourcecode/private/mx/impl/PropertiesWriter.cpp
webern: I think we need to adapt the `int alter` and `double cents` pattern wherever we convert to/from `alter double`. See #50
Probably what we should do is move the calculations that I did in #50 to the MxConverter thing and use it in both places.

CMakeLists.txt
webern: 👍

CMakeLists.txt
p-groarke: Sounds good, I'll rebase once everything is settled for the cmake stuff.

Sourcecode/private/mx/impl/PropertiesWriter.cpp
p-groarke: And by using a custom data struct instead of pitchdata, I'll use a double so it doesn't conflict.

p-groarke:
closed, fixed by https://github.com/webern/mx/pull/81

PR #82, opened 2020-04-02
make binary names lowercase

webern:
This change could be annoying on a case insensitive filesystem. But it seems somewhat uncool to name your binaries with uppercase letters.


webern:
Seems quite, don't think anyone is kicking the tires of mx at the moment, so I will merge this.

PR #90, opened 2020-05-08
Pitch data constructor

### LINKED ISSUE #69: [feature] Constructor with defaults.
Hi, I hope you are well and good amidst the current pandemic.

I've been hitting a minor snag when trying to store `mx::api` types in various const/global containers. It would be nice if `PitchData` offered a flexible value constructor to remedy this.

For example:
`PitchData(Step step = Step::c, int alter = 0, double cents = 0.0, Accidental c = Accidental::none, bool = false, bool = false, bool = false, bool = false, int octave = 4);`

This would enable a user to create pitches like this: `PitchData p{Step::c, -1};` which is nice.

A real world use case, I am building an `unordered_map` to map a root to a `mx::api::KeyData` fifth.

```c++
namespace mx {
using namespace api;
}

namespace std {
template <>
struct hash<mx::PitchData> {
	constexpr inline size_t operator()(const mx::PitchData& p) const {
		size_t h1 = size_t(p.step) << 32;
		size_t h2 = size_t(p.alter);
		return h1 ^ h2;
	}
};
} // namespace std


namespace {
std::unordered_map<mx::PitchData, int> root_to_key_circle {
    { mx::PitchData{/* you cannot do that :( */ }, -1 },
    { mx::PitchData{/* you cannot do that :( */ }, 0 },
    { mx::PitchData{/* you cannot do that :( */ }, 1 },
// etc
};
} // namespace
```

Now, you may want to have different or other types of constructors, but the general idea is the same. It would be nice to be able to construct a `PitchData` with a few arguments.

Take care and stay safe!

webern:
Closes #69

### DIFF
--- a/Sourcecode/include/mx/api/PitchData.h
+++ b/Sourcecode/include/mx/api/PitchData.h
@@ -69,6 +69,20 @@ namespace mx
             // default construction is middle c (c4)
             PitchData();
 
+            explicit PitchData(
+                Step inStep,
+                int inAlter = 0,
+                int inOctave = 4,
+                Accidental inAccidental = Accidental::none,
+
+                // less-often used params last
+                double inCents = 0.0,
+                bool inIsAccidentalParenthetical = false,
+                bool inIsAccidentalCautionary = false,
+                bool inIsAccidentalEditorial = false,
+                bool inIsAccidentalBracketed = false
+            );
+
             // the note name. i.e. c, d, e, f, g, a, b
             Step step;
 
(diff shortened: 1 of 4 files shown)


webern:
There is something horrible about `MSVC 19.25.28614.0` and I don't have time to chase it down. It's possible that there's undefined behavior in the `mx` code, but the last time I chased a Windows crash I ended up here: https://developercommunity.visualstudio.com/content/problem/847490/msvc-codegen-error-vector-reverse-iterator-x64-c17.html

The fix for mx was basically 'move something around so the buggy compiler doesn't make bad code'

#61 had the fix and it was here: https://github.com/webern/mx/blob/20796cd4638b8beec6f7833c436a7e267a5699c1/Sourcecode/private/mx/core/ElementInterface.cpp#L142...L150

PR #93, opened 2020-05-11
bump to C++17

webern:
Bumping to C++17 as `std::optional` in particular could make some things easier to express in `mx::api` instead of continuing the pattern of having a `bool isSpecified` for future optional things.


webern:
I need to figure out why GCC 10 is angry.

webern:
Looks like this may be the GCC 10 issue with C++20:
https://github.com/catchorg/Catch2/pull/1929/files

Issue #58, opened 2020-02-22
mx::core Support MusicXML 3.x


webern:
We are stuck on MusicXML 3.0 because it’s a big job to introduce upgrade to 3.1 (or whichever version is out when we get to this).

Do this without breaking mx::api

webern:
Pretty sure this is done, finally. After so many years. 🎉

PR #107, opened 2020-07-25
codegen: break up strings.h

webern:
Related to work on #58. Breaks up `Strings.h` such that there is a `.h` for each type.


webern:
> webern force-pushed the gen-strings.h-pr-2 branch from 63b20b4 to 7824950

Fix xcode project.

PR #118, opened 2021-01-26
readme: add new project setup example

### LINKED ISSUE #117: How exactly am I supposed to setup and use mx in a project?
Forgive me for the noob-ish question, but I tried following the setup guidelines you posted in `README.md` and was unable to actually use the library in the same manner as you did in the examples. In particular I followed your build instructions --

```
git clone https://github.com/webern/mx.git mx
mkdir build
cd build
cmake ../mx -DMX_BUILD_TESTS=on -DMX_BUILD_CORE_TESTS=off -DMX_BUILD_EXAMPLES=on
make -j6
./mxtest
```

And I also turned off all the CMake build options I only needed the library.

In the very first example, you used the below dependencies, which I tried compiling: (as a bare minimum example)

```
#include "mx/api/DocumentManager.h"
#include "mx/api/ScoreData.h"
```

Unfortunately my compiler couldn't find either file (assuming that my own program code is supposed to reside outside of MX's codebase)  so I tried specifying the direct path to both these files.

```
#include "/direct/path/to/mx/api/DocumentManager.h"
#include "/direct/path/to/mx/api/ScoreData.h"
```

The compiler was able to find both files, but they had internal relative references to other MX .h files, which the compiler also failed to find.

I'm 100% sure I'm doing something wrong, but I have no idea what it is. Any tips?

webern:
Closes #117 

Show in the readme how you can create a new project and use mx in-tree as a CMake subdirectory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,103 @@ You only need to run them if you make changes in the `mx::core` namespace.
 * `mx` third-party code should be checked into the `mx` repo and compiled as part of the `mx` library.
 * `mx` should not depened on any package manager, though it may be published into any package manager.
 
+### Using `mx` in a Cmake Project
+
+The following script demonstrates how you can start a new cmake project that uses `mx` by commiting its sourcecode into your project:
+
+```sh
+#!/bin/bash
+se -eou pipefail
+
+# this script demonstrates how to depend on mx by including it in your
+# sourcecode tree.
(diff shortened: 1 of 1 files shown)


webern:
Problem with xcode seems likely to be unrelated, created #119

PR #126, opened 2021-08-08
Add script to build XCFramework

SemitoneGene:
We don't have to assemble our own Fat binaries anymore using lipo.

Fixed a newline warning also.


webern:
Cool. Will look at this tomorrow.

PR #134, opened 2023-01-16
Add support for segno reading

SemitoneGene:
Segno are now read and made available in the public API.

### DIFF
--- a/src/private/mx/impl/DirectionWriter.cpp
+++ b/src/private/mx/impl/DirectionWriter.cpp
@@ -8,6 +8,7 @@
 #include "mx/core/generated/BassStep.h"
 #include "mx/core/generated/BeatUnitGroup.h"
 #include "mx/core/generated/Bracket.h"
+#include "mx/core/generated/Coda.h"
 #include "mx/core/generated/Dashes.h"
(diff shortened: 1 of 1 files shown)


SemitoneGene:
I also made a fix to catch.h in cpul that was causing a crash when building on the M1.

SemitoneGene:
Coda reading added as well.

SemitoneGene:
Added rehearsal reading.

In typical Gene fashion, I just realized all these changes only deal with reading. I am working on getting the writer work done.

SemitoneGene:
I stubbed in the writing functions, but need to do some more work to understand the correct way to write the info out. Any help is appreciated!

webern:
I'll take a look.

webern:
I think I fixed the git history at least. I force pushed to your fork and branch (because you can do that with GitHub, it's crazy).

So, Assuming your fork is at "origin", you should:

```sh
cd mx
git remote add upstream git@github.com:webern/mx.git
git checkout master
git fetch upstream master
git reset --hard upstream/master
git push origin -f master
git fetch origin gene/351
git checkout gene/351
git reset --hard upstream gene/351
```

PR #140, opened 2023-04-13
Export stops before starts

### LINKED ISSUE #139: Question about ordering of starts and stops for e.g. slurs
Currently, when I export a situation like this to music xml with `mx`
image
the output looks like
```
<notations>
   <slur type="start" number="1"/>
   <slur type="stop" number="1"/>
</notations>
```
Unfortunately, this breaks the slurs completely when I import it into MuseScore.
If I flip the ordering of the start and stop elements above, it works fine. 

Quoting from https://www.w3.org/2021/06/musicxml40/musicxml-reference/data-types/start-stop-continue/ this seems to be even required by the MusicXML standard itself:

> When multiple elements with the same tag are used within the same note, their order within the MusicXML document should match the musical score order. For example, a note that marks both the end of one slur and the start of a new slur should have the incoming slur element with a type of stop precede the outgoing slur element with a type of start.

I'd open a PR that changes the export code such that stops always precede starts when they are on the same note.

gergol:
Fixes #139

### DIFF
--- a/Sourcecode/private/mx/impl/DirectionWriter.cpp
+++ b/Sourcecode/private/mx/impl/DirectionWriter.cpp
@@ -158,6 +158,24 @@ namespace mx
                 }
             }
             
+            for( const auto& wedgeStop : myDirectionData.wedgeStops )
+            {
+                auto wedgeStartDirectionTypePtr = core::makeDirectionType();
+                wedgeStartDirectionTypePtr->setChoice( core::DirectionType::Choice::wedge );
+                this->addDirectionType( wedgeStartDirectionTypePtr, directionPtr );
+                auto wedgePtr = wedgeStartDirectionTypePtr->getWedge();
+                auto& attr = *wedgePtr->getAttributes();
+                attr.type = core::WedgeType::stop;
+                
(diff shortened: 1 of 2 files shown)


gergol:
The diff looks a bit weird, but I actually only flipped the ordering of the start/end code blocks.

webern:
Interesting. I'll take a closer look at this. Thank you.

webern:
I've created a test that fails *without* your changes. I'm going to try to add it to the PR.

```
/home/brigmatt/repos/mx/Sourcecode/private/mxtest/api/SlurTest.cpp:79: FAILED:
  CHECK( ("stop") == (stopValue) )
with expansion:
  "stop" == "start"

/home/brigmatt/repos/mx/Sourcecode/private/mxtest/api/SlurTest.cpp:84: FAILED:
  CHECK( ("start") == (startValue) )
with expansion:
  "start" == "stop"

deleting expected files................................done
creating expected files................................done
===============================================================================
test cases: 2672 | 2671 passed | 1 failed
assertions: 9878 | 9876 passed | 2 failed
```

gergol:
Thanks a lot!

webern (approved):

PR #143, opened 2023-06-25
Improve arpeggiate reading

SemitoneGene:
The Core::UpDown can't be used as the apreggiate reader will set the direction type to Up, which is not correct. The default arpeggiate does not have an arrow indicating up, so we need to introduce a "none" type.

### DIFF
--- a/Sourcecode/private/mx/core/Enums.cpp
+++ b/Sourcecode/private/mx/core/Enums.cpp
@@ -3830,6 +3830,42 @@ namespace mx
             return toStream( os, value );
         }
 
+        /// UpDownNone /////////////////////////////////////////////////////////////////////////////////
+
+        UpDownNone parseUpDownNone( const std::string& value )
+        {
+            const auto opt = tryParseUpDownNone( value );
+            return opt.value_or( UpDownNone::none );
+        }
+
+        std::optional<UpDownNone> tryParseUpDownNone( const std::string& value )
+        {
+            if( value == "up" ) { return UpDownNone::up; }
+            else if( value == "down" ) { return UpDownNone::down; }
+            return std::optional<UpDownNone>{};
+        }
+
+        std::string toString( const UpDownNone value )
+        {
+            switch ( value )
+            {
+                case UpDownNone::up: { return "up"; }
+                case UpDownNone::down: { return "down"; }
+                default: break;
+            }
+            return "up";
+        }
+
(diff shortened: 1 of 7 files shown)


Sourcecode/private/mx/impl/DirectionWriter.cpp
```diff
@@ -521,4 +521,4 @@ namespace mx
             }
         }
     }
-}
+}
```
webern, line 524: nit

```suggestion
}

```

Sourcecode/private/mx/impl/DirectionWriter.cpp
```diff
@@ -332,33 +332,6 @@ namespace mx
...
-            }
-
-            if( myDirectionData.rehearsals.size() > 0 ) {
-                
-                for( const auto& rehearsal : myDirectionData.rehearsals ) {
-                    auto outDirType = core::makeDirectionType();
-                    outDirType->setChoice( core::DirectionType::Choice::rehearsal );
-                    this->addDirectionType( outDirType, directionPtr );
-                }
-            }
```
webern: Are these code blocks no longer needed for some reason?

Sourcecode/private/mx/core/Enums.cpp
```diff
@@ -3830,6 +3830,42 @@ namespace mx
...
+            const auto opt = tryParseUpDownNone( value );
+            return opt.value_or( UpDownNone::none );
+        }
+
+        std::optional<UpDownNone> tryParseUpDownNone( const std::string& value )
+        {
+            if( value == "up" ) { return UpDownNone::up; }
+            else if( value == "down" ) { return UpDownNone::down; }
+            return std::optional<UpDownNone>{};
+        }
```
webern, line 3846: I think you can do away with the std:optional by changing these parsing functions to return UpDownNone::none in the default case. Or is there a place where we need to know the difference between optional::none and UpDownNone::none?

SemitoneGene: Sorry about the junk in the diff. I'll make a new one without the line ends and the sengo work that ended up in this diff.

I was afraid to dive in deeper to the use of the optional as other code uses the UpDown. The appregiate code could be smarter. As you state, it sets the default to Up, which add the direction attribute even if there was none present in the parsed XML.

PR #163, opened 2026-05-31
build: sdk toolchain container with ccache

webern:
## What

Replace the multi-stage `docker build` quality-gate model with a `docker run` SDK image plus ccache.

- build the pinned toolchain image (`mx-sdk`); rebuild only when `Dockerfile` changes.
- run gates with `docker run`.
- bind-mount the build dir for incremental build.
- use `ccache` as the CMake compiler launcher for better incremental builds.

## Why

Incremental build state persists across gate runs, and CI compiles become ccache hits across PRs
instead of cold every run.

## Notes

- Coverage folds into the run model: `gcovr` runs in the container and writes the report to the
  workspace mount.
- `--user` UID mapping on Linux keeps the cached ccache dir readable by the runner.
- Removed `Dockerfile.coverage` and `Dockerfile.coverage.dockerignore`.

Related to #157 which is tracked by #58.


README.md
```diff
@@ -53,6 +53,10 @@ switching modes never recompiles another mode's tree. Knobs: `JOBS` (parallelism
 
 A `core-dev` mode exists for codegen iteration on `mx/core`; see `AGENTS.md` for details.
 
+Certain aspects of the development workflow need to be reproducible regardless of host tooling
+(gating on warnings, for example). An `mx-sdk` docker image is used for this. See `Dockerfile` and
+`Makefile` for details on that.
```
webern:
```suggestion
`Makefile` for details.
```

PR #168, opened 2026-06-11
newgen

webern:


AGENTS.md
```diff
@@ -1,4 +1,215 @@
...
+baseline.
+
+### Flow (same in all three languages)
+
+1. **Discover** eligible `.xml`/`.musicxml` files under `data/`, excluding directories `expected`,
+   `testOutput`, `generalxml`, `smufl`, and files matching `*.fixup.xml` or having a `.invalid`
+   sibling marker.
+2. For each file:
+   a. Load the XML into a DOM.
+   b. Set the root `version` attribute to `"3.0"`.
```
webern: Seems wrong. Seems like this should be 3.0, 3.1 or 4.0

AGENTS.md
```diff
@@ -1,4 +1,215 @@
...
+  ```
+- `data/testOutput/corert/` - debug output directory for failure diffs (gitignored via build/).
+
+### Normalization pipeline
+
+Applied to both expected and actual documents before comparison:
+
+1. Set XML declaration: `<?xml version="1.0" encoding="UTF-8" standalone="no"?>`.
+2. Set DOCTYPE based on root element name (`score-timewise` vs `score-partwise`).
+3. Set root `version` attribute to `"3.0"`.
```
webern: Again, there are multiple versions in play in the repo and we will move this to 4.0

docs/sounds-4.1-0d56097.xml
```diff
@@ -0,0 +1,932 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE sounds PUBLIC "-//Recordare//DTD MusicXML 4.1 Sounds//EN" "http://www.musicxml.org/dtds/sounds.dtd">
+
+<!--
+	MusicXML sounds file
+
+	Version 4.1 Draft
```
webern: MusicXML 4.1 is not released and I believe the sounds version tracks 1:1 with the MusicXML version. Thus I believe this should be deleted.

data/lysuite/ly33d_Spanners_OctaveShifts.xml.invalid
```diff
@@ -0,0 +1,2 @@
+Not well-formed XML: the file begins with stray bytes ("Octave.cc") before the
```
webern: It's likely that I simply accidentally wrote Octave.cc at the beginning of this file. Just fix the test file by deleting the stray word instead of skipping it.

data/lysuite/ly75a_AccordionRegistrations.fixup.xml
```diff
@@ -4,18 +4,24 @@
...
     <value></value>
-    <replacement-value>0</replacement-value>
+    <replacement-value>1</replacement-value>
   </replace>
   <replace>
     <type>element</type>
     <name>accordion-middle</name>
     <value>test</value>
-    <replacement-value>0</replacement-value>
+    <replacement-value>1</replacement-value>
```
webern, line 13: Double check. Is this a NonNegativeInteger? Just want to make sure this correction is right.

data/synthetic/elision.3.0.xml
```diff
@@ -27,7 +27,7 @@
         <staff>1</staff>
         <lyric>
           <text>x</text>
-          <elision font-family="x" font-style="normal" font-size="1" font-weight="normal" color="#FF000000" underline="1" overline="1" line-through="1" rotation="1" letter-spacing="1" xml:lang="en" dir="ltr">x</elision>
```
webern, line 30: What happened here? Is the test file wrong? If these attributes exist in 3.0, then I would want them preserved in the test case for coverage. Are these attributes added in a later version of MusicXML? If so, do we have them for coverage in a 4.0 file?

data/synthetic/extend.3.0.xml
```diff
@@ -29,7 +29,7 @@
           <text>x</text>
           <elision>x</elision>
           <text>x</text>
-          <extend type="start" default-x="1" default-y="1" relative-x="1" relative-y="1" font-family="x" font-style="normal" font-size="1" font-weight="normal" color="#FF000000" />
```
webern, line 32: What happened here? Is the test file wrong? If these attributes exist in 3.0, then I would want them preserved in the test case for coverage. Are these attributes added in a later version of MusicXML? If so, do we have them for coverage in a 4.0 file?

gen/cpp/config.toml
```diff
@@ -0,0 +1,33 @@
...
+
+[sounds]
+# Companion sounds file (vendored under docs/, version-matched to the XSD).
+# Folds the standard instrument-sound identifiers into the IR as a sound-id
+# enum unioned with an open string. Comment out to disable.
+xml = "../../docs/sounds-4.0-ed15c23.xml"
+
+[target]
+language  = "cpp"
+namespace = "mx::core"
```
webern, line 29: Is this a "hardcoded" property, e.g. target.namespace is part of the prescribed structure for the toml file? If so, it seems a bit language-specific. Or, is it dynamic in the sense that I could say target.foo="bar"?

gen/emit/go/__init__.py
```diff
@@ -0,0 +1,115 @@
+"""Go backend: render the Plates into the Go test target package.
```
webern: Oh no. This is very bad and not what I wanted at all.

It should not be required to write a different python program for Go than for C. The whole point of the exercise was to make sure we did not have to write bespoke logic/code for specific targets. What is this doing and why is it needed?

We need to analyze what's being done in these bespoke Go and C backends and re-design so that these backends do not exist.

gen/plates/languages.py
```diff
@@ -0,0 +1,126 @@
...
+        "non_negative_integer": "int",
+    },
+}
+
+# How enum constants are scoped, per language: `bare` when the language
+# scopes them inside the type (C++ enum class), `composed` when they share
+# one flat namespace and carry the type's name (Go package-level constants,
+# C's single global namespace). This is a language fact, not configuration;
+# the composition itself happens in the projection so Variant.ident is final.
+VARIANT_SCOPES: dict[str, str] = {
```
webern: No. I do not want the generator to know about languages or have a list of languages that it supports.

webern:
Superseded by #169

PR #169, opened 2026-06-13
add support for musicxml 4.0 in mx::core

### LINKED ISSUE #157: Deshittify the 12,000 lines of reverse engineered python code in `gen/`

### LINKED ISSUE #158: Generate MusicXML 4.0 code

webern:
Initially I tried using AI to reverse-engineer the pseudo-hand-rolled original codegen. It worked but resulted in a 12k lines of unredeemable Python garbage with no room to maneuver from there. So, I started over and developed a `gen/` program from the ground up using AI. I tried to keep it agnostic as to language target, so I think targeting other languages and use-cases is a real possibility now. Early on, I used Go and C as language targets to force the AI to think about extensibility. Those targets exist under `gen/test`, but are intended more as `gen/` program regression tests than for actual MusicXML use.

For the replaced `mx::core` code, I prioritized compile time and better use of C++ features like `variant` and `option`. AI wanted to drop the `ezxml` abstraction and I guess it was time to let it go, so `pugixml` is promoted to `mx::core` interaction.

A `test-core-dev` target was used to allow the the AI to innovate on `mx::core` without worrying about `mx::impl` and `mx::api` (in fact, I deleted those layers during code-gen). Then I replaced those layers and burned tokens to preserve the `mx::impl` algorithms targeting the new set of `mx::core` classes.

## References

- Closes #157
- Closes #158
- Progresses #58
- Closes PR #167
- Closes PR #168

## Follow-ups:
- surface more features in the `mx/api` layer. top priority is probably SMuFL
- better packaging and distribution


BoseGene:
Let me look into the XCode target. What I have in my fork is an updated script to build a XCFramework, but it might use the workspace as the input.

PR #170, opened 2026-06-14
core: embed git SHA as build-time constant

webern:
## What

Adds `mx::core::gitSha` -- a C++ `const char* const` holding the git commit the library was built
from. Unclean trees get a `-dirty` suffix (any modified tracked file or any untracked file); builds
outside a git checkout (e.g. an exported source archive) get `"unknown"`.

## Why this shape

The capture must survive a rebuild that bypasses the Makefile. A configure-time capture
(`execute_process` / `configure_file`) freezes the SHA into the generated project and goes stale the
moment HEAD moves, because an IDE rebuild (Xcode/Ninja/Make) never re-runs CMake. So the capture is a
build step baked into the project (`cmake/GitSha.cmake`, run via `cmake -P`); it re-checks git every
build. `copy_if_different` rewrites the generated TU only when the value actually changed, so
incremental builds stay warm and only that one TU recompiles on a new SHA.

Declaration (`Version.h`) and definition (`Version.cpp.in` -> generated `GitSha.cpp` under the build
tree) are split so consumers that include the header never recompile on a SHA change.

## Files

- `src/private/mx/core/Version.h` -- the `gitSha` declaration
- `src/private/mx/core/Version.cpp.in` -- build-time template
- `cmake/GitSha.cmake` -- the capture + copy-if-different script
- `CMakeLists.txt` -- `mx_git_sha` custom target, wires the TU into `mx_core`

## Verification

- Full `mx_core` builds and links the symbol under the pinned GCC-14 toolchain
  (`_ZN2mx4core6gitShaE` in `libmx_core.a`).
- Native AppleClang: the generated TU compiles, links, and prints the SHA.
- Fallbacks confirmed: no git -> `unknown`; clean tree -> no suffix; untracked file alone ->
  `-dirty`.
- `copy_if_different` leaves the TU mtime unchanged when the SHA is unchanged (no needless rebuild).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,6 +45,8 @@ target_include_directories(pugixml PUBLIC ${PRIVATE_DIR})
 # own directory, so generator-owned code is recognizable at a glance) and
 # arrive via sources.cmake (emitted by `python3 -m gen gen/cpp/config.toml`).
 set(MX_CORE_RUNTIME_SOURCES
+    ${PRIVATE_DIR}/mx/core/Attribution.cpp
+    ${PRIVATE_DIR}/mx/core/Attribution.h
     ${PRIVATE_DIR}/mx/core/Decimal.cpp
     ${PRIVATE_DIR}/mx/core/Decimal.h
     ${PRIVATE_DIR}/mx/core/Error.h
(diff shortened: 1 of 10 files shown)


webern:
@copilot rebase

PR #235, opened 2026-06-21
test: api-level segno/coda round-trip coverage

webern:
Closes #204.

PR #203 added segno/coda support with impl-layer round-trip tests but deferred api-level coverage. This adds it through the public `DocumentManager` path, in two commits.

## Commit 1 — file-walking load coverage
Registers the audit fixtures `segno.{3.0,3.1}.xml` and `coda.{3.0,3.1}.xml` (under `data/synthetic/`) in `MxFileRepository::initializeNameSubdirectoryMap()` so the api file-walking suite (`ApiLoadSurvivalTest`) exercises the public read path (`createFromFile -> getData`) over them. The 3.1 variants additionally carry `id` and `smufl`; all four load and yield a part.

## Commit 2 — strict read→write→read gate
The synthetic probes cannot serve a strict DOM-equal gate: I dumped expected-vs-actual and confirmed the `<segno>`/`<coda>` elements round-trip byte-identically, but the probes also carry editorial `part-group`/`direction` children (`footnote`/`level`/`voice`) the api omits and have no notes (so the api synthesizes `<divisions>`) — divergences unrelated to segno/coda. The value-only `Fixer` can't patch structural diffs, and the existing real-world files (`testDalSegno.xml`, `testDCalCoda.xml`) fail on an unrelated `encoding-date` ordering issue.

So this adds two dedicated minimal fixtures that isolate the segno/coda surface (3.0 base attributes; 3.1 adds `id` + `smufl`) in a score that round-trips losslessly:

- `data/custom/segno_coda_roundtrip.3.0.xml`
- `data/custom/segno_coda_roundtrip.3.1.xml`

They are pinned in `roundtrip-baseline.txt` so the `test-api-roundtrip` CI gate enforces strict `createFromFile -> getData -> createFromScore -> writeToStream` DOM-equality for segno and coda, registered in `MxFileRepository`, and carry their audit `*.features.xml` sidecars.

## Testing
- `test-api-roundtrip` regression gate: **34 passed, 0 failed** (incl. both new fixtures).
- Full `mxtest`: **4268 assertions in 307 test cases, all pass**.
- `clang-format --dry-run --Werror` clean; audit sidecars generated via `python3 -m audit files` (no `corpus.xml` churn).

## References
- Closes #204
- Follows #203 (segno/coda support)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
_Generated by [Claude Code](https://claude.ai/code/session_0163ZKdf1FtkiVkpXG2K9C1s)_

Issue #228, opened 2026-06-20
api: round-trip injects implied default elements: `<type>` `<line>` `<voice>` `<staff>`


webern:
Surfaced by the reworked round-trip classifier (PR #225; part of #212).

## What

The api writer materializes elements the source left implicit, so the output has more of them than the input:

- `<type>` on notes that had none — extra `<type>` in `lysuite/ly41d_StaffGroups_Nested.xml`
- `<line>` under `<clef>` where the source omitted it — `lysuite/ly12a_Clefs.xml`
- `<voice>` where the source had none — `mjbsuite/ChordDirectionPlacement.xml` (expected: no `<voice>`; actual: `<voice>1</voice>`)
- `<staff>` where the source had none — `lysuite/ly02e_Rests_NoType.xml`

Related (same family, but never the sole blocker of a file): `add:divisions` (392) and `add:attributes` (359) — the writer also materializes implied containers.

## Impact (reorder-free candidate files; 550 total)

- `add:type` 233, `add:voice` 29, `add:line` 27, `add:staff` 1

## Decision

Per #214: decide per element whether the api should suppress these when the source omitted them, or whether injecting an implied default is acceptable and should be normalized away in comparison. This is the opposite direction from #219 (which tracks elements being dropped).

## Repro

```
make dump-api-roundtrip && make classify-api-roundtrip
# build/api/classified.json: signatures add:type / add:line / add:voice / add:staff
```

## References

- Part of #208; feeds #213
- Relates to #214 (divergence policy) and #219 (element drops)
- Surfaced by PR #225

Issue #221, opened 2026-06-20
track footnote and level support in api.features.xml


webern:
Surfaced by the round-trip failure classifier (#211, PR #217).

## What

`footnote` and `level` are each dropped in 373 failing files, but neither appears in `data/api.features.xml`. Because the classifier cross-references that index to decide whether a drop is "expected" (`support="none"`) or a bug (`support="full"/"partial"`), files whose only blockers are untracked elements fall into the `unknown` bucket — the classifier can't categorize what it can't look up. This is a measurement blind spot, not an mx::api code bug.

## What to do

Run the `api-feature-audit` skill to add `footnote` and `level` (and any other untracked-but-dropped elements it finds) to `data/api.features.xml` with the correct support level. Once tracked, the classifier can fold these files into B (drop-only, if support=none) or flag them as bugs (if support=full/partial), shrinking the `unknown` bucket.

## Repro

```
make dump-api-roundtrip && make classify-api-roundtrip 2>warnings.txt
grep "footnote" warnings.txt   # "unclassified FAIL (missing=[... 'footnote' ...])"
grep 'name="footnote"' data/api.features.xml   # no match
```

## References

- Part of #208
- Found by #211 / PR #217

webern:
Closing as resolved-by-removal rather than by tracking.

This issue assumed the fix was to add `footnote`/`level` to `data/api.features.xml` so the classifier could categorize the ~373 files that drop them. Investigating the editorial feature for #236 turned up the more fundamental fact: `<footnote>`/`<level>` appear in **zero** real-world corpus files. Every one of the ~382 editorial-bearing files was under `data/synthetic/`, where the audit generator stamps the boilerplate onto nearly every single-feature probe.

So the drops the classifier saw were noise, not a real api gap. Rather than model editorial in `mx::api` (the original #236 approach, now closed) or track a feature the wild corpus never uses, we removed the boilerplate at the source:

- #236 — closed; editorial is intentionally not modeled in `mx::api`
- #239 — strips `<footnote>`/`<level>` from every synthetic fixture except the dedicated `footnote`/`level` probes

With the boilerplate gone, those ~373 files no longer drop `footnote`/`level`, so the classifier blind spot this issue described is gone — without adding the untracked-support entry that was proposed here. The dedicated `footnote`/`level` probes still exist for spec coverage, and if we ever want the audit to record them as `support="none"` that's a trivial follow-up, but it's no longer blocking the round-trip classifier.

---
[...]

Issue #219, opened 2026-06-20
investigate api dropping elements marked support=full on round-trip


webern:
Surfaced by the round-trip failure classifier (#211, PR #217).

## What

The classifier diffs each failing file as an element multiset and cross-references `data/api.features.xml`. The most-dropped elements are almost all marked `support="full"` in our own audit, yet they vanish on round-trip:

| element | files affected | audited support |
|---|---|---|
| part-group | 373 | full |
| staff | 297 | full |
| lyric / text / syllabic | ~115 | full |
| page-margins + *-margin | ~99 | full |
| voice | 96 | full |
| measure-numbering | 88 | full |
| direction / direction-type | 71 | full |
| staff-layout | 33 | partial |

Spot-checked `synthetic/accent.3.0.xml`: expected has `<part-group>` twice, actual has it zero times.

## The question

Each of these is one of two things, and they need to be told apart per element:

1. a genuine impl round-trip bug (an element we claim to model is being dropped), or
2. `api.features.xml` overstating support (the element is really partial/none and the audit is wrong).

Either way it's actionable: (1) is a bug fix, (2) is an audit correction.

## Caveat

The comparison is strict full-DOM and this branch's baseline is ungrown (#208/#209), so these counts are the raw landscape, not a regression. #208 notes the api drops some subtrees by design — the point of this issue is the elements above are marked `full`, so by-design dropping would itself be the bug (a mislabel).

## Suggested approach

Triage the top offenders one at a time (start with part-group, staff, lyric, voice). The `add-mx-api-feature` skill covers the impl wiring; the `api-feature-audit` skill covers correcting support levels.

## Repro

```
make dump-api-roundtrip && make classify-api-roundtrip
# build/api/classified.json: missing_elements per file, cross-referenced to support
```

## References

- Part of #208
- Found by #211 / PR #217

webern:
Split into one issue per dropped element - each is triaged independently (genuine impl drop bug vs an `api.features.xml` support mislabel), since the elements are unrelated and likely need different handling:

- #274 - part-group
- #275 - staff
- #276 - lyric (with text/syllabic)
- #277 - page-margins (with the *-margin elements)
- #278 - voice
- #279 - measure-numbering
- #280 - direction (with direction-type)
- #281 - staff-layout

Closing as superseded by the above.

PR #236, opened 2026-06-22
feat: model editorial (footnote/level) in mx::api

webern:
## Summary

Models MusicXML editorial information (`<footnote>` and `<level>`) in `mx::api` so it survives an api round-trip instead of being silently dropped.

Adds a reusable `api::EditorialData` (`footnote` + `level`) and attaches it to:
- `PartGroupData` — editorial on the part-group start
- `DirectionData` — editorial on the `<direction>` (the editorial-voice `<voice>` is already represented by `DirectionData.voice`)

New api types: `EditorialData`, `FootnoteData` (formatted-text, modeled on `WordsData`), and `LevelData` with new `StartStopSingle` / `SymbolSize` enums. A shared `impl/EditorialFunctions.h` translates the core editorial groups (`EditorialGroup`, `EditorialVoiceDirectionGroup`) in both directions, and `Converter` gains the two enum bridge maps. `FootnoteData` reads color with the correct `isColorSpecified = color().has_value()` gate, avoiding the bug tracked in #207.

Part-group stop editorial is intentionally not modeled — MusicXML ignores child element values at a group stop.

## Scope

This makes footnote/level round-trip through the api, proven by unit tests. It does not pin the dense `synthetic/{segno,coda}.*` probes to the strict round-trip baseline: after this change those files still diverge on two unrelated gaps — the writer materializing `<divisions>` in a note-less measure (the #228 family) and a dropped direction `<voice>` — tracked in a separate follow-up.

## Testing

- [x] New `EditorialApiTest`: footnote+level round-trip on a part-group and a direction, with distinct non-default values per field (29 assertions, 2 cases)
- [x] Full `mxtest`: 4297 assertions in 309 test cases, all pass
- [x] api round-trip regression gate: 34/34 (no regressions)
- [x] `clang-format` clean

## Notes

Stacked on #235 (base branch `claude/issue-204-plan-vsvcas`); review/merge that first. GitHub will retarget this to `main` once #235 merges.

## References

- Follows #235 / #204 (segno/coda api round-trip)
- Progresses #221 (footnote/level support; the `api.features.xml` audit update is the natural next step) and #219 (part-group/direction children dropped on round-trip)
- Part of #208 (api round-trip coverage)
- Uses the color pattern from #207

---
_Generated by [Claude Code](https://claude.ai/code/session_0163ZKdf1FtkiVkpXG2K9C1s)_

### DIFF
--- a/src/include/mx/api/LevelData.h
+++ b/src/include/mx/api/LevelData.h
@@ -1,0 +1,66 @@
+// MusicXML Class Library
+// Copyright (c) by Matthew James Briggs
+// Distributed under the MIT License
+
+#pragma once
+
+#include "mx/api/ApiCommon.h"
+
+#include <string>
+
+namespace mx
+{
+namespace api
+{
+// Mirrors the MusicXML <level> 'type' attribute (start-stop-single): whether the editorial
+// information applies to the start of a series of symbols, the end, or a single symbol.
+enum class StartStopSingle
+{
+    unspecified,
+    start,
+    stop,
+    single
+};
+
+// Mirrors the MusicXML <level> 'size' attribute (symbol-size).
+enum class SymbolSize
+{
+    unspecified,
+    full,
+    cue,
+    graceCue,
+    large
+};
+
+// The MusicXML <level> element specifies editorial information for the parent element. Its text
+// content is descriptive; the attributes control how the editorial marking is rendered. An
+// `unspecified` enum (or empty `value`) means the source carried no such attribute and none is
+// written back.
+class LevelData
+{
+  public:
+    std::string value;
+    Bool reference;
+    StartStopSingle type;
+    Bool parentheses;
+    Bool bracket;
+    SymbolSize size;
+
+    LevelData()
+        : value{}, reference{Bool::unspecified}, type{StartStopSingle::unspecified}, parentheses{Bool::unspecified},
+          bracket{Bool::unspecified}, size{SymbolSize::unspecified}
+    {
+    }
+};
+
+MXAPI_EQUALS_BEGIN(LevelData)
+MXAPI_EQUALS_MEMBER(value)
+MXAPI_EQUALS_MEMBER(reference)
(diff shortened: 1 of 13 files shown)


src/include/mx/api/LevelData.h
```diff
@@ -0,0 +1,66 @@
...
+// information applies to the start of a series of symbols, the end, or a single symbol.
+enum class StartStopSingle
+{
+    unspecified,
+    start,
+    stop,
+    single
+};
+
+// Mirrors the MusicXML <level> 'size' attribute (symbol-size).
```
webern, line 25: This strikes me as one of things we should try to simplify in the `mx::api` model instead of mirroring the way MusicXML models it. In general, we are trying to get away from MusicXML's procedural "stop"/"start" semantics (i.e. requiring stateful parsing) and represent data more statically whenever we can.

webern:
Closing without merging — we're dropping editorial (`<footnote>`/`<level>`) from `mx::api` entirely.

Rationale: editorial appears in **zero** real-world corpus files. Every `<footnote>`/`<level>` in `data/` (386 files) is in `synthetic/`, where the audit generator stamps boilerplate editorial onto nearly every single-feature probe. So this PR was adding api surface — including the start/stop/single `<level>` `type` machinery — purely to satisfy synthetic boilerplate, not anything in the wild. We don't want the synthetic corpus to drive features that real files never use.

What happens to the pieces of this PR:
- **Direction `<voice>`** (the genuinely useful part, and the actual blocker for pinning the synthetic segno/coda fixtures) moves to #238, decoupled from the editorial scaffolding. `<voice>` *is* common in the wild.
- **The synthetic fixtures** get their editorial boilerplate stripped instead: #238 cleans the four segno/coda fixtures it pins; a follow-up PR strips footnote/level from every other synthetic file that isn't the dedicated `footnote`/`level` probe.

Superseded by #238 (now based directly on #235).

---

Issue #237, opened 2026-06-22
api: segno/coda synthetic probes blocked from strict round-trip by empty-measure `<divisions>` and dropped direction `<voice>`


webern:
## Resolution

Closing as a duplicate of #228. The two blockers split cleanly:

- Blocker 2 (dropped direction `<voice>`) is fixed in #238 — the editorial-voice `<voice>` on a `<direction>` now round-trips.
- The four synthetic fixtures `synthetic/{segno,coda}.{3.0,3.1}.xml` are pinned to the strict gate in #238 (after removing the meaningless part-group stop editorial and giving each a real note so its `<divisions>` is legitimate).
- Blocker 1 — the writer materializing `<attributes>`/`<divisions>` into a note-less measure — is the writer-policy bug already tracked by #228 (the `add:divisions`/`add:attributes` family). A naive "suppress when the part has no notes" guard regresses `ksuite/k016a` (which legitimately carries `<divisions>` in a note-less measure), so the correct fix needs divisions-presence tracking — #228's call, not a spot fix here.

Nothing actionable remains in this issue beyond #228, so it is repointed there.

---

## Original report

Follow-up to #204 / #235 (segno/coda api coverage) and #236 (editorial footnote/level support).

### The two blockers

1. Empty-measure `<attributes>`/`<divisions>` injection. The synthetic measure has no notes, so the writer materializes a `<divisions>` the source never had — the writer-materializes-implied-containers family of #228 (`add:divisions` / `add:attributes`).

2. Dropped direction `<voice>`. The `<direction>` carries `<voice>1</voice>` (editorial-voice-direction group), which the api dropped on write. Fixed in #238.

### Repro

```
cmake --build build/api --target mxtest-api-roundtrip
./build/api/mxtest-api-roundtrip discovery data --dump /tmp/dump
diff /tmp/dump/synthetic__segno.3.0.xml.expected.xml /tmp/dump/synthetic__segno.3.0.xml.actual.xml
```

### References

- Follows #204 / #235 / #236; voice blocker fixed in #238
- Divisions blocker is the #228 writer-policy family
- Part of #208 (api round-trip coverage)

Issue #204, opened 2026-06-17
api: add api-level round-trip test for segno/coda synthetic fixtures


webern:
PR #203 added segno/coda support and writes impl-layer round-trip tests, but the api-level test was deferred.

## Task

Register `segno.3.0.xml`, `segno.3.1.xml`, `coda.3.0.xml`, and `coda.3.1.xml` in `MxFileRepository::initializeNameSubdirectoryMap()` (they live under `data/synthetic/`) so the file-walking api test suite exercises the full `DocumentManager` read→write→read path, not just the impl layer.

The 3.1 variants add `id` and `smufl`; both variants must pass. All four fixtures are already on disk — no new files needed.

## References

- PR #203 (segno and coda support — deferred this task)

PR #238, opened 2026-06-22
feat: round-trip direction voice and pin the synthetic segno/coda fixtures

webern:
Brings the audit's synthetic `segno`/`coda` fixtures onto the strict api read -> write -> read gate.

## What

1. **Round-trip the editorial-voice `<voice>` on a `<direction>`** (read from `editorialVoiceDirection().voice()`, written back). Today a direction's own `<voice>` is dropped. The write is emitted only when the direction has direction-type content, so a voice carried in from a figured-bass/harmony element never produces an empty `<direction>`. `<voice>` is common in the wild corpus, so this is real fidelity, not synthetic-only.

2. **Pin `synthetic/{segno,coda}.{3.0,3.1}.xml`** in `roundtrip-baseline.txt`. To make them round-trip losslessly the fixtures are tidied:
   - stripped the `<footnote>`/`<level>` editorial boilerplate the audit generator stamps onto every probe (editorial appears in **zero** real-world files, so the api does not model it — see #236, closed)
   - added a `<divisions>` + a `<note>` so the writer's first-measure `<divisions>` is legitimate instead of a spurious injection into a note-less measure

   They still isolate the segno/coda surface (3.0 base attributes; 3.1 adds id + smufl) and now also exercise a part-group and direction `<voice>`/`<staff>` through the public `DocumentManager` path.

## Why a real note instead of an impl fix for `<divisions>`

A note-less measure that the writer fills with a synthesized `<divisions>` is the #228 writer-policy issue. A naive "suppress divisions when the part has no notes" guard regresses `ksuite/k016a_Miscellaneous_Fields.xml`, which legitimately carries `<divisions>` in a note-less measure — distinguishing "source had it" from "api invented it" needs divisions-presence tracking in the public api, out of scope here. So the fixtures carry a real note; the writer-policy gap stays open under #228.

## Follow-up

A separate PR on top of this one strips the editorial boilerplate from every *other* synthetic file that isn't the dedicated `footnote`/`level` probe, so the synthetic corpus stops implying api features the wild corpus never uses.

## Testing

- [x] api round-trip regression gate: 38 passed, 0 failed (was 34; +4 synthetic)
- [x] discovery: no regressions
- [x] Full `mxtest`: 4268 assertions in 307 test cases, all pass
- [x] corert: 832 test cases pass, pinned corpus count unchanged
- [x] `clang-format` clean; audit `*.features.xml` sidecars + `corpus.xml` regenerated

## References

- Stacked on #235 / #204 (segno/coda api round-trip)
- Supersedes #236 (editorial dropped); resolves the direction-voice part of #237, with the `<divisions>` writer-policy part staying with #228
- Part of #208 (api round-trip coverage)

### DIFF
--- a/src/private/mx/impl/DirectionWriter.cpp
+++ b/src/private/mx/impl/DirectionWriter.cpp
@@ -21,6 +21,7 @@
 #include "mx/core/generated/DirectionTypeChoiceChoice.h"
 #include "mx/core/generated/Divisions.h"
 #include "mx/core/generated/Dynamics.h"
+#include "mx/core/generated/EditorialVoiceDirectionGroup.h"
 #include "mx/core/generated/Empty.h"
 #include "mx/core/generated/Figure.h"
 #include "mx/core/generated/FiguredBass.h"
@@ -110,6 +111,17 @@ std::vector<core::MusicDataChoice> DirectionWriter::getDirectionLikeThings()
         coreOffset.setValue(core::Divisions{core::Decimal{static_cast<double>(offset)}});
         coreOffset.setSound(core::YesNo::yes());
         direction.setOffset(coreOffset);
+    }
+
+    // Assign the direction to a voice via the editorial-voice <voice>. Only emitted when this
+    // direction also writes direction-type content (guarded by myIsFirstDirectionTypeAdded below),
+    // so a voice carried in from a figured-bass or harmony element never produces an empty
+    // <direction>.
+    if (myDirectionData.voice >= 0)
+    {
+        core::EditorialVoiceDirectionGroup editorialVoice{};
+        editorialVoice.setVoice(std::to_string(myDirectionData.voice));
+        direction.setEditorialVoiceDirection(std::move(editorialVoice));
     }
 
     for (auto mark : myDirectionData.marks)
--- a/src/private/mxtest/api/roundtrip-baseline.txt
+++ b/src/private/mxtest/api/roundtrip-baseline.txt
@@ -65,3 +65,11 @@ lysuite/ly72b_TransposingInstruments_Full.xml
 # round-trips losslessly.
 custom/segno_coda_roundtrip.3.0.xml
 custom/segno_coda_roundtrip.3.1.xml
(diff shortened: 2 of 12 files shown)


src/private/mx/impl/DirectionWriter.cpp
```diff
@@ -113,11 +113,26 @@ std::vector<core::MusicDataChoice> DirectionWriter::getDirectionLikeThings()
         direction.setOffset(coreOffset);
     }
 
-    if (myDirectionData.editorial.isFootnoteSpecified || myDirectionData.editorial.isLevelSpecified)
+    // Editorial (<footnote>/<level>) and the editorial-voice <voice> on the <direction>. This only
+    // reaches the wire when the direction has direction-type content (see myIsFirstDirectionTypeAdded
+    // below); a figured-bass/harmony-derived voice never produces a stray <direction>.
```
webern: This comment is hard to understand, try again. Also "wire" is not easy to understand here. Don't use "wire" be specific. Serialization? Translation? What are you saying.

src/private/mxtest/api/roundtrip-baseline.txt
```diff
@@ -60,12 +60,18 @@ lysuite/ly71a_Chordnames.xml
...
-# them through the strict DOM compare for reasons unrelated to segno/coda.
-# These two fixtures isolate the segno/coda surface (3.0 base attributes; 3.1
-# adds id + smufl) in a minimal score that round-trips losslessly.
+# Dedicated segno/coda api read->write->read fixtures (#204), isolating the
+# segno/coda surface (3.0 base attributes; 3.1 adds id + smufl) in a minimal
+# score that round-trips losslessly.
 custom/segno_coda_roundtrip.3.0.xml
 custom/segno_coda_roundtrip.3.1.xml
+
+# The audit's synthetic single-feature segno/coda fixtures now round-trip too,
```
webern: "now round trip too".

This is a static code comment, not a news headline. Stateless please.

Issue #214, opened 2026-06-19
api: define divergence policy and guardrails for round-trip corpus coverage


webern:
Sub-task of #208. See the parent tracking issue for the full plan and context.

Scope: keep discovery non-gating, decide per divergence class whether it belongs in the api, in a fixup-style expected alteration, or is out of scope, and optionally track a near-miss count to catch drift.

webern:
The #219 round-trip drop batch (#274-#281) resolved to a concrete divergence policy per file, matching this issue's decision framework:

- Genuine impl bugs, fixed: #279 measure-numbering (#322), #275 staff (#323), #281 staff-layout (#326)
- api design choice incompatible with strict round-trip fidelity, absorbed into the test harness rather than changed (documented, deliberate): #277 page-margins odd/even-\>both collapse (#327)
- Not a bug at all -- already correctly audited/resolved: #274 part-group (closed directly), #278 voice / #280 direction (#325, documents why `data/api.features.xml` already correctly marks the specific unmodeled `direction-type` subtypes as `support="none"`; only the parent container is `full`, correctly)
- Out of scope, tracked separately as real feature gaps rather than folded into a quick fix: #324 (unmodeled `direction-type` subtypes and pedal states)

Closing this phase-5 batch now that PRs are open for the #219 sub-issues; the policy above is the concrete precedent for future divergence classification.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #213, opened 2026-06-19
api: implement ranked fixes and grow the round-trip corpus baseline


webern:
Sub-task of #208. See the parent tracking issue for the full plan and context.

Scope: implement the ranked fixes one at a time; after each, re-run discovery, confirm targeted files flip to `PASS` with no regressions, append their paths to `src/private/mxtest/api/roundtrip-baseline.txt`, and commit.

webern:
## Ranked round-trip fix worklist (from PR #225)

Measured over 550 reorder-free candidate files (`make dump-api-roundtrip && make classify-api-roundtrip` → `build/api/classified.json`). Ordered by the greedy batch plan (minimal changes → most files). The first two issues land ~26 files outright; all 15 underlying fixes land 71.

| # | fix | key signatures | candidate files |
|---|-----|----------------|-----------------|
| #226 | stop injecting `<identification>`/`<encoding>` | add:encoding (535), add:identification (420) | ~535 |
| #227 | preserve `part-name`/`part-abbreviation` @print-object | attr:part-name@print-object (532) | ~532 |
| #228 | stop injecting implied defaults (note type, clef line, voice, staff) | add:type (233), add:voice (29), add:line (27) | ~280 |
| #230 | keep modeled attributes (clef @number, repeat @times, cancel @location, credit-words @justify) | attr:clef@number (29) | ~35 |
| #229 | keep zero-valued fields (clef-octave-change, diatonic, stem @default-y) | drop:clef-octave-change (14) | ~20 |
| #231 | keep accidental editorial attrs (cautionary/editorial/parentheses) | attr:accidental@editorial (5) | ~9 |

Distance-1 instant wins (one fix → PASS): `attr:stem@default-y` (#229), `drop:diatonic` (#229), and three `add:encoding`-only files (#226).

Already tracked separately, not re-filed: element drops — staff/voice/lyric/text/etc. (#219), encoding-child reorder (#220), footnote/level (#221). Reorders (278 files) are deferred to divergence policy (#214).

---
_Generated by [Claude Code](https://claude.ai/code)_

webern:
The round-trip drop issues split out of #219 (#274-#281) now all have PRs open:

- #274 (part-group) -- investigated, already resolved by earlier work, closed directly (no PR needed)
- #275 (staff) -- #323
- #277 (page-margins) -- #327
- #278 (voice) -- #325
- #279 (measure-numbering) -- #322
[...]

PR #243, opened 2026-06-26
feat: round-trip <encoding> faithfully and add a writeMxVersion toggle

### LINKED ISSUE #220: api reorders identification/encoding children on round-trip
Surfaced by the round-trip failure classifier (#211, PR #217).

## What

51 files are classified "reorder-only" (category C): the output contains exactly the same elements as the input (the tag multiset is identical, nothing dropped or added) but in a different order. Every one of them diverges at the same place:

```
/score-partwise/identification/encoding/software
```

So the api is writing the children of `<encoding>` (software, encoding-date, supports, ...) in a different order than the source file. The music is fully intact; the file just isn't a byte-faithful copy.

## Fix options

1. Make the api writer emit `<encoding>` children in MusicXML schema order, or
2. Treat encoding-child order as a non-difference in comparison normalization (this overlaps with the divergence policy in #214).

Option 1 is the more faithful fix; option 2 is the cheaper one if encoding order is deemed insignificant.

## Severity

Low — content is preserved, only ordering differs. But it currently blocks 51 files from the strict pass-list.

## Repro

```
make dump-api-roundtrip && make classify-api-roundtrip
# build/api/classified.json: primary_category "C", first_divergence_path on encoding
```

## References

- Part of #208
- Relates to #214 (divergence policy)
- Found by #211 / PR #217

webern:
## Summary

Resolves the `<encoding>` child-order round-trip divergence (#220) and adds a way to suppress mx's own provenance stamp. Three parts:

1. Encoding child order is treated as insignificant in the api round-trip. `mx::api` flattens `<encoding>` into typed buckets (`EncodingData`) and re-emits its children in a fixed canonical order, so files that list those children in any other order round-tripped with an identical multiset but a different order — the sole blocker for a batch of files. The api round-trip harness now stable-sorts `<encoding>` children into the api's canonical order on both sides before comparing; corert stays order-faithful and is untouched. This also realigns the user `<software>` against mx's trailing stamp, clearing a secondary spurious value mismatch. Pins the 48 files this unblocks (discovery PASS 46 to 94).

2. The writer now emits `<software>` before `<encoding-date>`, matching what MuseScore, Finale, and Sibelius write, so fewer real-world files need any reordering. MusicXML's encoding choice imposes no order, so this is schema-valid.

3. New `api::EncodingData::writeMxVersion` (defaults true, including for files parsed without the stamp). When set false, `writeToFile`/`writeToStream` serialize without mx's provenance `<software>` while still writing the user's own software entries. The directive cannot ride on the core model, so `DocumentManager` carries it per document (`StoredDocument`) and branches at write time.

## Testing

- [x] `make test-api-roundtrip` — 94 / 94 pinned files pass (was 46)
- [x] `make test` — full unit suite (4275 assertions in 309 cases), including 2 new `writeMxVersion` round-trip tests (default-on stamps; off suppresses)
- [x] `make discover-api-roundtrip` — 94 PASS / 737 FAIL, no regressions from the writer reorder
- [x] `make check` — clang-format clean

## References

- Closes #220
- Progresses #213 (grow the pinned baseline) and #208
- Relates to #214 (divergence policy — encoding order resolved as option 2)

### DIFF
--- a/src/include/mx/api/EncodingData.h
+++ b/src/include/mx/api/EncodingData.h
@@ -60,6 +60,11 @@ class EncodingData
     std::vector<std::string> software;
     std::vector<SupportedItem> supportedItems;
     std::vector<MiscellaneousField> miscelaneousFields;
+
+    // When true (the default, including for files parsed without it), mx writes its
+    // own provenance <software> node into <encoding> on every api write. Set false
+    // to suppress that stamp; the user's own <software> entries are still written.
+    bool writeMxVersion = true;
 };
 
 MXAPI_EQUALS_BEGIN(SupportedItem)
@@ -90,6 +95,7 @@ MXAPI_EQUALS_MEMBER(encodingDescription)
 MXAPI_EQUALS_MEMBER(software)
 MXAPI_EQUALS_MEMBER(supportedItems)
 MXAPI_EQUALS_MEMBER(miscelaneousFields)
+MXAPI_EQUALS_MEMBER(writeMxVersion)
 MXAPI_EQUALS_END;
 MXAPI_NOT_EQUALS_AND_VECTORS(EncodingData);
 } // namespace api
--- a/src/include/mx/api/EncodingData.h
+++ b/src/include/mx/api/EncodingData.h
@@ -61,9 +61,17 @@ class EncodingData
     std::vector<SupportedItem> supportedItems;
     std::vector<MiscellaneousField> miscelaneousFields;
 
-    // When true (the default, including for files parsed without it), mx writes its
-    // own provenance <software> node into <encoding> on every api write. Set false
-    // to suppress that stamp; the user's own <software> entries are still written.
+    // mx stamps its own provenance <software> node (the mx URL plus the build's git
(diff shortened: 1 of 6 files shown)


src/include/mx/api/EncodingData.h
```diff
@@ -60,6 +60,11 @@ class EncodingData
     std::vector<std::string> software;
     std::vector<SupportedItem> supportedItems;
     std::vector<MiscellaneousField> miscelaneousFields;
+
+    // When true (the default, including for files parsed without it), mx writes its
```
webern: This should explain that the intent is to aid in fixing mx defects. In other words, since this is a bit of an intrusive defect we should include our reason for doing so.

webern: Expanded the comment in 97fb7a2 to give the rationale: the stamp is deliberately intrusive (it injects a `<software>` node the source never had into every written file), it exists to tie bad output back to the exact mx build that produced it, and this flag is the escape hatch for byte-faithful output when diagnosing mx defects.

---

PR #244, opened 2026-06-26
build: allow clients to provide their own pugixml

### LINKED ISSUE #242: Pugixml conflict
I am reviewing MX potentially to use as an object model for a new musicxml export from Finale musx files. (This would be an MIT open source standalone CLI exporter that does not require Finale.) MX, especially `mx::api` seems like a perfect fit for my goals, which are to create a semantic export that does not try to capture engraving details that don't work on a consuming platform anyway and more likely just get in the way. The MX object model seems like it will make the code far easier to develop, understand, and maintain.

One of the issues I am concerned about is that this repo uses `pugixml` and [my project](https://github.com/rpatters1/denigma) also uses `pugixml`. Because MX uses a vendored copy, there is a potential conflict with my repo that gets it with FetchContent. The LLM is telling me the one here could be protected with a cmake namespace or just a different name, like `mx_pugixml`.

I could post LLM-generated cmake suggestions here. (I mostly vibe code cmake, sad to say.) But I am curious how open you would be to making a change.

If MX ever adds automated `mxl` support we could have the same problem with `mini_zip` or whatever zip tool you decide to vendor in. Reading `musx` files requires both `zip` and `xml` handling, so my project has both. I take to heart your proscription of dependencies in the README. I hate dependencies, too. But vendoring rather than getting them with FetchContent means extra care is required to support client projects using the same dependencies.

rpatters1:
This PR gates the inclusion of the local vendored `pugixml` on whether another already exists. It allows a client project that also uses `pugixml` to include `mx` without creating a conflict.

Resolves #242

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,7 +39,7 @@ set(PUBLIC_DIR "${CMAKE_CURRENT_SOURCE_DIR}/src/include")
 # Client projects may provide a compatible pugixml target before adding mx.
 if(NOT TARGET pugixml)
     add_library(pugixml STATIC ${PRIVATE_DIR}/pugixml/pugixml.cpp)
-    target_include_directories(pugixml PUBLIC ${PRIVATE_DIR})
+    target_include_directories(pugixml PUBLIC ${PRIVATE_DIR}/pugixml)
 endif()
 
 # mx_core: the generated C++ typed model plus its hand-written runtime
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,8 +36,11 @@ set(PUBLIC_DIR "${CMAKE_CURRENT_SOURCE_DIR}/src/include")
 
 # pugixml: the XML layer the generated mx::core consumes directly (plan §2.1).
 # Vendored at src/private/pugixml (upstream 1.15, stock pugiconfig).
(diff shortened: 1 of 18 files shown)


webern (commented):
Thanks for working on this. As written it won’t quite work because I have dumb header paths for the pugi headers. There’s a small fix needed in the Cmake file and the include paths need to change at all the sites where pugi is referenced.

I’m on a mobile phone right now. Later I can add a commit to this PR or you can do it if you get there first:

```bash
git remote add upstream https://github.com/webern/mx.git
git fetch upstream pugi-headers
git cherry-pick 99dc0a9
```

Also the CI failures are something I need to fix. It looks like my comment posting workflows for code coverage don’t work for anyone but me.

rpatters1:
It was working for me, weirdly, so I never looked at the `include` sites. Thank you for the quick response. I updated the PR with your change.

webern:
> It was working for me, weirdly, so I never looked at the `include` sites. Thank you for the quick response. I updated the PR with your change.

I think it worked because your version of pugixml and the vendored version were the same, or at least close enough that the headers did not differ in any binary-affecting way. So I think when you were building mx you were reading the vendored headers but linking the cpp's you had built outside.

Anyway, I had forgotten to update one thing so did a force push to that third commit (mine). And checking CI 👍

webern (approved):
Nice, thank you for fixing this!

PR #246, opened 2026-06-27
gen: replace anonymous namespace with per-type name in enum impls

### LINKED ISSUE #245: Support for unity builds
Given the large number of generated classes in `mx::core`, I suspect it would greatly speed up builds to be able to use unity builds on it.

I have experimented, and unity builds seem to work on other subsystems, such as `mx::api`. But they fail on `mx::core` because many if not all translation units have an anonymous namespace containing same-named items.

It seems like a fairly simple fix to the Python generator to change these anonymous namespaces to something like

```cpp
namespace generated_detail::<ClassName> {

}
```

Then prepend that namespace as needed. With this change, unity builds would be possible.

What are the ramifications of doing this? Is it something you would consider?

webern:
## Summary

Each generated enum \`.cpp\` file (136 of them) contained an anonymous \`namespace {}\` wrapping a
\`constexpr std::string_view kWire[]\` lookup table. In a normal build this is fine — each
translation unit gets its own anonymous namespace and the names never conflict. In a unity build,
however, all included \`.cpp\` files share one translation unit, so the anonymous namespaces merge
and 136 definitions of \`kWire\` collide.

The fix is minimal: drop the anonymous namespace and give the array a per-type name
(\`kYesNoWire\`, \`kAboveBelowWire\`, etc.) so the names are unique within \`mx::core\` and unity
builds can include multiple enum files without redefinition errors.

Changes:
- `gen/cpp/templates/enum.cpp.tmpl` — remove the anonymous namespace, rename `kWire` to `k{{ident}}Wire`
- `src/private/mx/core/generated/*.cpp` — 97 regenerated files (the enum-class strategy files that had the anonymous namespace)

## Testing

- [ ] `make test` passes

## References

- Closes #245

### DIFF
--- a/gen/cpp/templates/enum.cpp.tmpl
+++ b/gen/cpp/templates/enum.cpp.tmpl
@@ -8,16 +8,11 @@
 namespace {{vars.namespace}}
 {
 
-namespace
-{
-
-constexpr std::string_view kWire[] = {
+constexpr std::string_view k{{ident}}Wire[] = {
 {{#variants}}
     {{name.wire_q}},
 {{/variants}}
 };
-
-} // namespace
 
 {{#variants}}
 {{type.ident}} {{type.ident}}::{{ident}}() noexcept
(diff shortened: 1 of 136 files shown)


rpatters1:
I left this comment at the issue as well: we should probably do the same for the 5 hand-crafted mx::core TUs that have anonymous namespaces. With that change, I can unity build the entire MX subsystem and it is *much* faster and less noisy.

PR #267, opened 2026-06-29
fix(api): clef number round-trip

### LINKED ISSUE #270: api: clef number attribute does not round-trip
Split out of #230, which grouped unrelated attribute drops on different elements. One repro per issue.

## What

The clef `number` attribute does not round-trip. Investigation (PR #267) found the divergence is the
reverse of a plain drop: for single-staff parts mx *added* a spurious `number="1"` on mid-measure
clef changes, where the source omits it (1 is implied). The classifier signature `attr:clef@number`
covers any mismatch in either direction.

Verified on `lysuite/ly46c_Midmeasure_Clef.xml`.

## Impact

`attr:clef@number` — 29 reorder-free candidate files (the largest single attribute signature in the worklist).

## Repro

```
make dump-api-roundtrip && make classify-api-roundtrip   # signature attr:clef@number
```

## References

- Split from #230 · Part of #208 · feeds #213 · surfaced by PR #225
- Fixed by PR #267

webern:
## Summary

Round-trip added a spurious `number="1"` to clefs in single-staff parts. The issue frames this as a
dropped attribute, but for the affected corpus files it is the reverse: the source omits the clef
number (1 is implied on a single staff) and mx emitted it anyway on mid-measure clef changes. The
measure-start path already suppressed it; the two mid-measure write paths passed the staff index
straight through.

The fix centralizes the single- vs multi-staff rule in `MeasureWriter::clefStaffIndex()` so all
three clef write sites agree, and adds `ClefData::writeStaffNumber` - a ternary where `unspecified`
applies the rule automatically (omit the implied 1 on a single-staff part, include otherwise) and
`yes`/`no` force it. Reading sets `yes`/`no` only when the source diverges from the rule, so it
stays `unspecified` in the common case and most users can ignore it.

Also includes a tooling fix: `make dump-api-roundtrip` now clears its output directory first. It
only writes pairs for failing files, so a file that flipped to passing left a stale pair that the
classifier re-counted as a current failure.

## Testing

- [x] New `MultiStaffClefKeepsNumber` test; `PropertiesButNoNotes` updated (it pinned the old bug)
- [x] 23 clef files flip FAIL -> PASS, added to the round-trip baseline (96 -> 119 pinned)
- [x] Regression gate green (119 of 119 pinned)
- [x] Full impl/api suite passes (4330 assertions in 319 test cases)

## References

- Closes #270 (clef `number` round-trip)
- Progresses #213 (grows the round-trip corpus baseline by 23 files)

### DIFF
--- a/src/include/mx/api/ClefData.h
+++ b/src/include/mx/api/ClefData.h
@@ -38,7 +38,12 @@ class ClefData
   public:
     ClefData();
 
-    // int staffIndex;
+    // Most users can ignore this; leave it unspecified. It only controls whether the clef's
+    // optional number attribute is written. unspecified (the default) applies the right rule
+    // automatically: omit the number on a single-staff part (where 1 is implied) and include it
+    // otherwise. yes/no force the attribute on or off. It exists for round-trip fidelity - reading
+    // a file sets yes/no only when the source diverged from the automatic rule.
+    Bool writeStaffNumber;
     ClefSymbol symbol;
     int line;
     // When true (the default), the writer emits <line>. Set to false when the source
@@ -73,7 +78,7 @@ class ClefData
 };
 
 MXAPI_EQUALS_BEGIN(ClefData)
-// MXAPI_EQUALS_MEMBER( staffIndex )
+MXAPI_EQUALS_MEMBER(writeStaffNumber)
 MXAPI_EQUALS_MEMBER(symbol)
 MXAPI_EQUALS_MEMBER(line)
 MXAPI_EQUALS_MEMBER(isLineSpecified)
--- a/src/private/mxtest/api/roundtrip-baseline.txt
+++ b/src/private/mxtest/api/roundtrip-baseline.txt
@@ -155,3 +155,32 @@ musuite/testWords1.xml
 # and NoteWriter dropped valid zero values on round-trip.
 custom/transposition.musicxml
 ksuite/k002a_Fermatas.xml
+
+# Unblocked by #230 (clef @number round-trip). The writer emitted a spurious
+# number="1" on mid-measure clef changes in single-staff parts (the staffIndex
+# was passed straight through instead of being suppressed for single-staff, as
+# the measure-start path already did), and did not preserve a source-specified
+# number. Source clefs here carry no number, so 1 is assumed and must be omitted.
+ksuite/k006a_Clefs_sign_G_ln_2_oct_0_Treble.xml
+ksuite/k006b_Clefs_sign_F_ln_4_oct_0_Bass.xml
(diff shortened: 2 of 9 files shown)


src/include/mx/api/ClefData.h
```diff
@@ -38,7 +38,15 @@ class ClefData
   public:
     ClefData();
 
-    // int staffIndex;
+    // The staff this clef belongs to (1-based). Populated on read; 1 for a single-staff part.
+    // Building via the API can ignore it - the writer derives the staff from position.
```
webern: This is not very clear. Do you mean that a user, when building with the API, need not specify this and it will be ignored? Unclear.  Should users leave it set to the default value of -1 and the API just takes care of it if so? Does the API ignore it only if it is 0 or -1, or does the API take it and use it if it is provably incorrect? How does this work?

src/include/mx/api/ClefData.h
```diff
@@ -38,7 +38,15 @@ class ClefData
...
+    int staffNumber;
+    // Most users can ignore this; leave it unspecified. It only controls whether the clef's
```
webern:
```suggestion
    int staffNumber;

    // Most users can ignore this; leave it unspecified. It only controls whether the clef's
```

src/private/mxtest/impl/MeasureWriterTest.cpp
```diff
@@ -178,14 +178,54 @@ TEST(PropertiesButNoNotes, MeasureWriter)
     CHECK_DOUBLES_EQUAL(101.0, props.divisions()->value().value(), 0.01);
 
     CHECK_EQUAL(1, props.clef().size());
-    CHECK(props.clef().back().number().has_value());
-    CHECK(1 == props.clef().back().number()->value());
+    // Single-staff part (numStaves == 1): the clef carries no number; 1 is assumed. The
+    // mid-measure write path previously emitted a spurious number="1" (#230a).
```
webern:
```suggestion
    // mid-measure write path previously emitted a spurious number="1" (#230).
```

src/private/mxtest/api/roundtrip-baseline.txt
```diff
@@ -155,3 +155,32 @@ musuite/testWords1.xml
 # and NoteWriter dropped valid zero values on round-trip.
 custom/transposition.musicxml
 ksuite/k002a_Fermatas.xml
+
+# Unblocked by #230a (clef @number round-trip). The writer emitted a spurious
```
webern:
```suggestion
# Unblocked by #230 (clef @number round-trip). The writer emitted a spurious
```

PR #283, opened 2026-06-30
Add `<staff-details><staff-size>` to API.

rpatters1:


src/private/mx/impl/PropertiesWriter.cpp
```diff
@@ -190,9 +195,19 @@ void PropertiesWriter::writeStaffDetails(int staffIndex, int staffLines)
...
-    sdg.setStaffLines(staffLines);
-    staffDetails.setGroup(sdg);
+    if (staffLines >= 0)
+    {
+        core::StaffDetailsGroup sdg{};
+        sdg.setStaffLines(staffLines);
+        staffDetails.setGroup(sdg);
+    }
+
+    if (staffSize >= 0.0)
```
webern, line 210: Is >= correct (and not >)? Is zero valid and would we want to write it?

webern (commented):
Seems fine. Just eyed it and left one question.

src/private/mx/impl/PropertiesWriter.cpp
rpatters1: Good question. ~I'm going to assume 0 is *not* valid and make changes accordingly. My question is whether we should use 0.0 as the sentinel rather than a negative value.~

rpatters1: LLM says:

>In MusicXML, <staff-size> is a non-negative-decimal, so 0 is schema-valid. Semantically it is odd: the element is a percentage of default staff scaling, so normal values are around 100, smaller staves might be something like 75, and 0 would mean a zero-height staff space. But because the schema allows non-negative values, the API sentinel should stay negative, and staffSize >= 0.0 is the right “specified” check.

rpatters1: We could apply a stricter policy if you wish. It wouldn't bother me.

webern: It's weird but oh well. That's MusicXML!

Issue #249, opened 2026-06-28
api: replace value + `isSpecified` bool pairs with an explicit optional value type


webern:
## Problem

`mx::api` round-trips optional-with-default fields by pairing a value with a parallel "specified" bool:

- `ClefData::line` + `isLineSpecified` (#228)
- `ClefData::octaveChange` + `isOctaveChangeSpecified` (#229)
- `TransposeData::diatonic` + `isDiatonicSpecified` (#229)
- the static-data analog `TimeSignatureData::isImplicit`

The value and the bool are two sources of truth kept in sync by hand. Forget the bool and the field silently fails to round-trip (or, for default-`true` flags, appears unbidden). It only bites the authoring path — the reader always sets the bool — so it is a pure public-API tax. It is most awkward for event-like data (`StaffData::clefs` is a vector; the object's existence already means "emit this," yet a flag can suppress it) and more defensible for static-per-measure data like `timeSignature`. **Breaking:** it changes public `ScoreData` field types.

## Options

**Solution A — `std::optional<T>`** — intrinsic presence, but `.value()` throws and `operator*` is UB on empty; safe defaulting needs `value_or` discipline.

**Solution B — `std::variant<std::monostate, T>`** — Rust-`Option`-like with `std::visit` exhaustiveness, but heavy for one payload and `std::get` still throws. (Right tool for multi-case choices like `AttributesChoice`, not a lone int.)

**Solution C — custom fused type (recommended)** — value + presence in one object; reads never throw, writes set both atomically:

```cpp
template <typename T>
class Optional // name TBD: clashes with std::optional
{
    T myValue{};
    bool myIsSpecified = false;

  public:
    bool isSpecified() const { return myIsSpecified; } // drives the writer
    T get() const { return myValue; }                  // always safe; default when unset
    void set(T v) { myValue = v; myIsSpecified = true; }
    void clear() { myValue = T{}; myIsSpecified = false; }
};
```

Solution C is the only one giving both intrinsic presence *and* safe default reads.

### Naming

`Optional` is a working name only — **TBD**, since it clashes with `std::optional` and behaves differently (defaults instead of throwing). The final name should signal the defaulting behavior (`Defaulted` / `Specifiable` / `Tracked`).

## Scope

Migrate #229 fields first (`octaveChange`, `diatonic`), then #228 (`isLineSpecified`, `isDurationNameSpecified`), then reconcile `isImplicit`. Preserve round-trip fidelity; update accessors and `MXAPI_EQUALS`.

[...]

PR #286, opened 2026-07-01
feat(api): support lv (let-ring) ties

### LINKED ISSUE #284: Support for LV ties
I'm wondering about the best way to support LV ties in the api. Would it make sense to add `isTieLetRing` to `NoteData`? It would have somehow be exclusive with `isTieStart`, if only by convention. (That is, only one could be set.)

Another approach would be to add an API-level enum for tie starts that had 2 values (TieStart::regular, TieStart::letRing). But this would be a breaking change.

webern:
## Summary

Adds support for laissez-vibrer (lv, or "let-ring") ties to mx::api.

An lv tie applies to a single note and has no matching stop, so it does not fit the existing
start/stop tie model. Rather than overloading `NoteData::isTieStart` / `isTieStop`, it is modeled
with its own start-only structure, `api::TieLetRing` (the shared visual attributes: `PositionData`,
`CurveOrientation`, and optional `ColorData`), held in `NoteData::tieLetRing` as a
`std::optional<TieLetRing>`. In MusicXML this is a lone `<tied type="let-ring">` notation; the
sound-level `<tie>` element has no let-ring type.

- Reader: `parseCurve` in `CurveFunctions.h` routes a lone `<tied type="let-ring">`
  (`core::TiedType::letRing`) into `NoteData::tieLetRing`. An `if constexpr` guard keeps the branch
  out of the slur instantiation, whose type has no let-ring alternative.
- Writer: `NotationsWriter` emits a `<tied type="let-ring">` from the field, carrying position,
  orientation, and color.

## Testing

- [x] Local build and tests were NOT run by the author. Relying on CI.
- Added unit tests in `CurveFunctionsTest.cpp`: `parseTieLetRing_*` (position, placement, color,
  orientation), `parseCurve_letRing_routesToTieLetRing`, `parseCurve_start_doesNotSetTieLetRing`,
  `writeAttributesFromTieLetRing_*`, and `tieLetRing_roundTrip`.

## References

- Closes #284

### DIFF
--- a/src/include/mx/api/NoteData.h
+++ b/src/include/mx/api/NoteData.h
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include "mx/api/CurveData.h"
 #include "mx/api/DurationData.h"
 #include "mx/api/LyricData.h"
 #include "mx/api/NoteAttachmentData.h"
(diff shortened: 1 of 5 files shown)


src/include/mx/api/NoteData.h
```diff
@@ -75,6 +76,43 @@ enum class Stem
...
+// tie, which is a start/stop pair joining two notes of the same pitch, an lv
+// tie applies to a single note and has no matching stop: it instructs the
+// renderer to let the note ring undamped. In MusicXML this is a lone
+// <tied type="let-ring"> notation (there is no sound-level <tie> counterpart,
+// since the <tie> element's type is only start or stop). Because it is
+// structurally unlike a start/stop tie, it is modeled with its own struct
+// rather than by overloading NoteData's isTieStart / isTieStop.
+struct TieLetRing
+{
+    // Whether this note carries an lv (let-ring) tie. When false the remaining
```
webern: This doesn't quite make sense to me. We might need this, not so sure, but if we do maybe isLvSpecified. It seems like this would be cleaner without this boot at all.

The weirder part is why this would govern all the other attributes.

src/include/mx/api/NoteData.h
```diff
@@ -98,6 +136,10 @@ class NoteData
     bool isTieStart;
     bool isTieStop;
 
+    // A laissez-vibrer / let-ring tie on this note (a lone <tied type="let-ring">
+    // with no matching stop). Independent of isTieStart / isTieStop. See TieLetRing.
+    TieLetRing tieLetRing;
```
webern: Oh. I see why we have that bool above. Hmm. This is dumb. I think we need to either use std::optional for the first time or consider the IsSpecified/Optional class I opened an issue about. #249

webern (commented):
Not mergeable as-is.

src/include/mx/api/NoteData.h
rpatters1: In my experience, `std::optional` is a great (arguably the best) solution to this kind of problem for any type except an `enum` or a `bool`. For those types, `std::optional` is codesmell. Instead of `optional`, enums should provide an `unspecified` option (which should be the `{}` initialization default) and bools should be converted to 3-state enums (which this repo already does).

rpatters1:
I'm wondering if we should create a similar class like `ContinuationTie` that models ties at the start of 2nd endings and the like.  Not for this PR but something to think about.

PR #301, opened 2026-07-05
feat: model grace and cue as independent bools on NoteData

### LINKED ISSUE #288: Grace notes inside cues
The current `mx::api` allows notes to be normal, grace, or cue. But that leaves grace notes inside cues out. Rather than an enum, would it not make more sense to simply have two booleans `isGrace` and `isCue`?

The other option I can see is to add a 4th option to the enum, but it feels like an enum is the wrong model for this.

webern:
## Human Summary

This seems ok, I guess, for now. We re-model `grace` and `cue` notes as bool flags on `NoteData`, even though the spec is a little weird (perhaps with good reason) in that `grace` cannot have duration and `cue` cannot have a tie (that's the one that seems really weird to me). So when reading and writing from MusicXML there is no problem.

The only issue is that when writing hand-authored `NoteData`, an author does not have any way to know that these would be dropped on write. It seems better than not supporting `grace` and `cue` combinations, which is what the issue was about, so this seems like a good-enough "for now" fix to me.

## Summary

MusicXML permits four `<note>` flavors: normal, cue, grace, and grace-cue (a grace note inside a cue passage). The api modeled this as a three-value enum (`NoteType`), which cannot represent grace-cue, so a `<grace/><cue/>` note round-tripped as grace-only: `NoteReader` never set the cue bit for core's `GraceCueNoteGroup` and `NoteWriter` had no branch emitting it.

Per the discussion on the issue, `NoteData` now carries two independent bools, `isGrace` and `isCue`, whose four combinations map one-to-one onto the four note flavors.

**This is a BREAKING api change** (removes `enum class mx::api::NoteType` and the `NoteData::noteType` field), intended for the batched breaking release. Migration: `noteType == NoteType::grace` becomes `isGrace`, `noteType == NoteType::cue` becomes `isCue`, `noteType == NoteType::normal` becomes `!isGrace && !isCue`.

Details:

- `NoteData`: enum deleted; the new bools carry doc comments stating the schema facts (grace notes have no wire `<duration>` -- ticks read as 0 and are ignored on write; cue and grace-cue notes cannot carry `<tie>` -- ties on them are silently dropped on write, which preserves the existing behavior for cue notes). Equality block and constructor defaults updated.
- `NoteReader`: sets the cue flag when a grace note group's choice is `GraceCueNoteGroup`.
- `NoteFunctions`: assigns the two bools directly instead of collapsing to an enum.
- `NoteWriter`: branches on the bool pair -- (F,F) `NormalNoteGroup`, (F,T) `CueNoteGroup`, (T,F) `GraceNoteGroup`/`GraceNormalNoteGroup`, (T,T) `GraceNoteGroup`/`GraceCueNoteGroup`.
- Corpus: new `data/synthetic/grace-cue.4.0.xml` (validates against the 4.0 XSD) exercising all four combinations in one measure; corert pinned count bumped to 833; audit artifacts regenerated (this also backfilled the sidecar for `data/rpatters1/timesigs_composite-ref.musicxml`, which #287 added without one); the fixture is pinned in `roundtrip-baseline.txt` since it survives the strict DOM compare.

## Testing

- [x] New `GraceCueApiTest` covers all four combinations through the api round-trip, the reader path (embedded `<grace/><cue/>` fixture), exact writer output (grace before cue, no `<duration>` on grace notes), tie dropping on cue and grace-cue, and tie preservation on grace-normal notes
- [x] `make test`: all pass (4497 assertions in 351 test cases, plus the three examples)
- [x] `make test-core-dev`: all pass (834 test cases; pinned count 833, 0 skipped)
- [x] `make test-api-roundtrip`: 122 passed, 0 failed (of 122 pinned, including the new `synthetic/grace-cue.4.0.xml`)
- [x] `xmllint --nonet --schema docs/musicxml-4.0-ed15c23.xsd data/synthetic/grace-cue.4.0.xml` validates

## References

- Closes #288


src/include/mx/api/NoteData.h
```diff
@@ -127,7 +120,14 @@ class NoteData
     // with no matching stop). Independent of isTieStart / isTieStop. See TieLetRing.
     std::optional<TieLetRing> tieLetRing;
 
-    NoteType noteType; // normal, cue, grace
+    // MusicXML's four <note> flavors -- normal, grace, cue, and grace-cue --
+    // are the four combinations of these two independent flags. Schema facts:
+    // a grace note carries no <duration> on the wire (durationTimeTicks reads
+    // as 0 and is ignored on write), and cue notes -- including grace-cue
+    // notes -- cannot carry <tie> (ties on them are silently dropped on write).
```
webern, line 127: Checked this against the XSD directly (`docs/musicxml-4.0-ed15c23.xsd`, the `note` complexType). Its own doc annotation says it outright: "grace notes do not have a duration element... Cue notes have a duration element... but no tie elements." Confirmed on the `mx::core` side too: `NormalNoteGroup`/`GraceNormalNoteGroup` have a `tie()` accessor, `CueNoteGroup`/`GraceCueNoteGroup` don't. So this is describing a real schema constraint, not something `mx::api` is choosing to impose.

---

src/private/mx/impl/NoteWriter.cpp
```diff
@@ -264,25 +265,10 @@ void NoteWriter::setNoteChoiceAndFullNoteGroup(bool isStartOfChord) const
 {
     myOutFullNoteGroup.setChord(myCursor.isChordActive && myIsPreviousNoteAChordMember && !isStartOfChord);
 
-    switch (myNoteData.noteType)
+    // The schema has no <tie> on cue and grace-cue notes, so ties on them are
+    // silently dropped. Normal and grace-normal notes carry their ties.
+    if (!myNoteData.isCue)
```
webern, line 270: Worth being explicit about the gap this leaves. Round-tripping is safe (see the `NoteReader.cpp` comment), but nothing stops an API consumer from authoring a `NoteData` with `isCue=true` plus `isTieStart`/`isTieStop` set, or `isGrace=true` plus a nonzero `durationTimeTicks` -- those get silently dropped here (and the grace-duration case a few lines down in `assembleNoteChoice`) with no exception or signal. That mirrors the pre-existing behavior for cue-note ties (previously handled by the old `NoteType` enum's `case cue: break;`), just now extended consistently to grace-cue. Given there's no validation layer elsewhere in the writer either, I think that's the right call for the api's current shape. Flagging as a comment, not asking for a change.

---

src/private/mx/impl/NoteReader.cpp
```diff
@@ -212,6 +212,12 @@ void NoteReader::setNormalGraceCueItems()
         {
             setTie(noteGuts.graceNoteChoice().asGraceNormalNoteGroup().tie());
         }
+        else
+        {
+            // <grace/> + <cue/>: a grace note inside a cue passage. The
+            // grace-cue group carries no <tie> in the schema.
+            myIsCue = true;
+        }
```
webern, line 220: Traced this against `NoteWriter`'s `assembleNoteChoice`/`setNoteChoiceAndFullNoteGroup`. Round-tripping is lossless by construction here, not just by luck: this function can only ever set `myDurationValue = 0` for a grace note (a few lines up, unconditional -- `GraceNoteGroup`'s core type has no `duration()` to read), and can only ever leave tie unset for cue/grace-cue -- as here, where the grace-cue branch sets `myIsCue` but never calls `setTie()`, since `CueNoteGroup`/`GraceCueNoteGroup` have no `tie()` to read from. So the four combinations this reader can produce are exactly the four the writer can re-emit -- a legitimately-parsed file can't lose data on round trip.

---

webern (commented):
Spot-checked the grace/cue split against the XSD and traced both the reader and writer paths. Three notes inline: the "schema has no &lt;tie&gt;" claim checks out against the actual spec text, round-tripping is lossless by construction (not just by observed test results), and the authoring-time silent-drop for schema-incompatible combinations is a deliberate, pre-existing policy this PR extends consistently rather than a new gap. Comment only, not blocking.

---

PR #303, opened 2026-07-05
CLOSE-me: add per-staff and composite time signatures to mx::api

### LINKED ISSUE #268: api: time signature improvements
The api should permit multiple time sigs per part, similar to what it does now for key signatures. And each time signature should support multiple beats/beat-type pairs per time signature.

This is a breaking change, so I won't submit a PR without feedback, but I want track it.

webern:
## Human Summary

Ha, well I asked the machine to pick low hanging fruit from @rpatters1's issues that it could "just do" without my supervisions since I am working a lot for my employer right now, and for some reason it thought these two were easy and small.

I don't love the time signature change, to be honest, I might not want to merge this with that time-signature solution as is. I'd like for there to be a normal path for the 99% of time signatures that are just a X over Y, and you don't have this ugly weird vector in the 99% case. So I think this might need to be reworked.

## Summary

`mx::core` already fully modeled the composite/staff-scoped time-signature shape (`Time::number()`, `TimeChoiceGroup::timeSignature()` as `OneOrMore`); the gap was entirely in `mx::api`/`mx::impl`.

- `TimeSignatureData` replaces the single `beats`/`beatType` strings with `std::vector<TimeSignatureComponent>` (`components`), supporting composite/interchangeable meters (multiple beats/beat-type pairs in one `<time>`). A `setSimple(beats, beatType)` helper covers the common single-pair case.
- `TimeSignatureData` gains `staffIndex` (same `INDEX_UNSPECIFIED` convention as `KeyData::staffIndex`) for staff-scoped `<time number="...">`.
- `MeasureData::timeSignature` (singular) becomes `timeSignatures` (a vector), carried forward per staff across measures the same way the old singular field carried forward via `isImplicit`. An unscoped `<time>` declaration replaces the whole effective list; staff-scoped declarations merge into the carried-forward list by `staffIndex`.
- `TimeReader` now parses every `<time>` element in a measure's `<attributes>` (previously only the first) and every beats/beat-type pair within each (previously only the first).

Reference corpus file `data/rpatters1/timesigs_composite-ref.musicxml` (added by #287) exercises the composite-meter case end to end.

## Testing

- [x] New tests: staff-scoped and composite-meter time signature round-trips, `setSimple()`, equality coverage for the new fields (`TimeSignatureApiTest.cpp`)
- [x] `make test`: all pass (4587 assertions in 362 test cases, plus the three examples)
- [x] `make test-api-roundtrip`: 124 passed, 0 failed (of 124 pinned; no regressions)

## References

- Closes #268

### DIFF
--- a/src/include/mx/api/MeasureData.h
+++ b/src/include/mx/api/MeasureData.h
@@ -31,7 +31,11 @@ class MeasureData
     // this is the notes and other music data
     std::vector<StaffData> staves;
 
-    TimeSignatureData timeSignature;
+    // resolved, carried-forward time signature(s) in effect for this measure. Normally a single
+    // entry with staffIndex == INDEX_UNSPECIFIED (applies to all staves). When staff-scoped time
+    // signatures are used, there is one entry per staff. Mirrors the isImplicit carry-forward
+    // semantics of the old singular field.
+    std::vector<TimeSignatureData> timeSignatures;
 
     // an empty measureNumber indicates a normal
     // measure number (i.e. the measure's
@@ -100,7 +104,7 @@ class MeasureData
     std::vector<TransposeData> transpositions;
 
     MeasureData()
-        : staves{}, timeSignature{}, number{}, measureNumbering{MeasureNumbering::unspecified},
+        : staves{}, timeSignatures{TimeSignatureData{}}, number{}, measureNumbering{MeasureNumbering::unspecified},
           multiMeasureRest{VALUE_UNSPECIFIED}, implicit{Bool::unspecified}, nonControlling{Bool::unspecified},
           width{DOUBLE_UNSPECIFIED}
     {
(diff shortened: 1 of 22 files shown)


rpatters1:
Maybe in addition to an easy setter there could be an easy getter routine that returns `std::optional<std::pair<...>>`. That way there would be no ugly math: it would get the simple time if it is simple and return std::nullopt if it isn't.

edit: or it could simply be

```cpp
bool getSimple(std::string& beats, std::string& beatType);
```

Where a return value of false means it isn't simple, but either way it returns the first entry.

rpatters1:
Or (and this is more what I think you want): keep beats and beatType as they are and emit them first, followed by the others in the array, if any. I'm not sure what to call the array. Maybe `secondaryTimeSignatures`. From a data modeling perspective it is wrong, but from an api usability perspective it might be a reasonable compromise.

webern:
Closing in favor of #321

PR #309, opened 2026-07-06
fix: round-trip the copyright rights type attribute

webern:
## Human Summary

A surprising fruitful unlock of `mx::api` round trip test files just by adding a missing `type` attribute to the `copyright` element.

## Summary

`ScoreData::copyright` is a plain string with no way to say whether the source's `<rights>`
declared a `type` attribute at all. `ScoreWriter` force-wrote `type="copyright"` unconditionally, so
a source `<rights>Public Domain</rights>` (no type) round-tripped as
`<rights type="copyright">Public Domain</rights>` -- a spurious attribute addition (`attr:rights@type`).

Added `ScoreData::copyrightType` (`optional<string>`, defaulting to `"copyright"` so authoring a
score from scratch is unchanged -- fidelity knobs default to automatic per the api design
principles). `ScoreReader` now records the selected `<rights>`'s actual type (or unset if it had
none); `ScoreWriter` omits the attribute when unset instead of always writing `"copyright"`.

The reader's selection criteria are unchanged: only a `<rights>` typed `"copyright"` (or untyped)
populates this field, matching the existing `composer`/`lyricist` creator-type handling. Other
rights types (e.g. `"mechanical"`) can be *written* via the new field but are out of scope for the
read side, same as the pre-existing `composer`/`lyricist` fields.

Stacked on `#308` (lyric syllabic fix).

## Testing

- [x] New tests (`DocumentManagerTest.cpp`): default-is-copyright, explicit-value writes verbatim
      (write-only, since the reader doesn't recapture arbitrary types), and unset omits the attribute
- [x] Existing `copyright` / `RoundTrip_copyright_0` tests still pass unchanged
- [x] `make test`: all pass (4574 assertions in 362 test cases, plus the three examples)
- [x] `make test-api-roundtrip`: 150 passed, 0 failed (of 150 pinned; 3 newly unlocked)

## References

- No tracking issue existed for this gap; found via corpus round-trip discovery/classification.

### DIFF
--- a/src/include/mx/api/ScoreData.h
+++ b/src/include/mx/api/ScoreData.h
@@ -16,6 +16,7 @@
 #include "mx/api/SystemData.h"
 
 #include <map>
+#include <optional>
 #include <vector>
 
 namespace mx
@@ -61,6 +62,14 @@ class ScoreData
     std::string arranger;
     std::string publisher;
     std::string copyright;
+
+    // The <rights> type attribute for `copyright` above. unset -> the source's <rights>
+    // had no type attribute (omitted on write); set -> written back verbatim, defaulting
+    // to "copyright" when authoring a new copyright string from scratch. On read, only a
+    // <rights> typed "copyright" (or untyped) populates `copyright`/`copyrightType` at
+    // all -- other rights types are out of scope for this field, matching `composer`/
+    // `lyricist` above, which likewise only recognize one fixed creator type each.
+    std::optional<std::string> copyrightType;
     EncodingData encoding;
     std::vector<PageTextData> pageTextItems;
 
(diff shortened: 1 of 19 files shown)


src/include/mx/api/ScoreData.h
```diff
@@ -61,6 +62,14 @@ class ScoreData
...
     std::string publisher;
     std::string copyright;
+
+    // The <rights> type attribute for `copyright` above. unset -> the source's <rights>
+    // had no type attribute (omitted on write); set -> written back verbatim, defaulting
+    // to "copyright" when authoring a new copyright string from scratch. On read, only a
+    // <rights> typed "copyright" (or untyped) populates `copyright`/`copyrightType` at
+    // all -- other rights types are out of scope for this field, matching `composer`/
+    // `lyricist` above, which likewise only recognize one fixed creator type each.
+    std::optional<std::string> copyrightType;
```
webern:
```suggestion
    // The <rights> type attribute for `copyright` (above).
    std::optional<std::string> copyrightType;
```

Issue #312, opened 2026-07-06
fix: consecutive grace notes at the same tick collapse on round-trip


webern:
## Summary

Two (or more) consecutive grace notes sharing the same `tickTimePosition`, followed by the normal
note they attach to, collapse to a single note on `createFromScore` -> `writeToStream` ->
`createFromStream` -> `getData`. Confirmed independent of any other feature (reproduces with plain
grace notes, no slash/tie/other attributes involved).

## Repro

```cpp
ScoreData score;
score.ticksPerQuarter = 4;
score.parts.emplace_back();
auto &part = score.parts.back();
part.measures.emplace_back();
auto &measure = part.measures.back();
measure.staves.emplace_back();
auto &staff = measure.staves.back();
auto &voice = staff.voices[0];

voice.notes.emplace_back();
voice.notes.back().isGrace = true;
voice.notes.back().tickTimePosition = 0;
voice.notes.back().durationData.durationName = DurationName::eighth;

voice.notes.emplace_back();
voice.notes.back().isGrace = true;
voice.notes.back().tickTimePosition = 0;
voice.notes.back().durationData.durationName = DurationName::eighth;

voice.notes.emplace_back();
voice.notes.back().tickTimePosition = 0;
voice.notes.back().durationData.durationName = DurationName::quarter;
voice.notes.back().durationData.durationTimeTicks = 4;

// round trip via createFromScore -> writeToStream -> createFromStream -> getData
// expected: 3 notes back. actual: 1 note back.
```

The write side looks correct (both `<grace/>` notes serialize with a `<backup>` between them to
reset the cursor), so the bug is on the read side -- something in the grace-note/backup handling
during measure parsing drops the earlier grace note(s) when more than one shares a tick.

## How this was found

Found while implementing grace note `slash` attribute support (`#312`ish, see the `grace-note-slash`
branch/PR): a first draft of that PR's test used two consecutive grace notes at the same tick and
hit this collapse. The final PR test avoids the scenario (one grace note per case), so this bug is
independent and unpinned by that change.

## Suggested next step

Bisect where the note count drops -- likely in the core roundtrip's cursor/backup handling for
grace notes (`MeasureReader`/`MeasureWriter` or the cursor advance logic), not the api layer
specifically, since the serialized XML itself is correct.

PR #313, opened 2026-07-06
feat: round-trip grace note slash attribute

webern:
## Human Summary

Simple fix for a missing attribute.

## Summary

`<grace>`'s `slash` attribute (slashed grace notes, e.g. acciaccatura) was parsed by nothing and
written by nothing -- `NoteData` had no field for it (`attr:grace@slash`).

Added `NoteData::graceSlash` (tri-state `Bool`, only meaningful when `isGrace` is true).
`NoteReader` reads it off the grace note's `<grace>` element, mirroring the existing
`getIsStemSpecified`/`getStem` accessor pattern already used for other optional grace-adjacent
attributes; `NoteWriter` emits `<grace slash="..."/>` only when specified, otherwise the bare
`<grace/>`.

`stealTimePrevious`/`stealTimeFollowing`/`makeTime` (the other three `<grace>` attributes) remain
unmodeled -- no corpus file currently needs them.

While writing this PR's test, an unrelated pre-existing bug surfaced: two consecutive grace notes
sharing a tick position collapse to one note on round-trip, independent of `slash`. Filed as #312
rather than folded into this fix; the test here avoids that scenario (one grace note per case).

Stacked on `#311` (strong-accent direction fix).

## Testing

- [x] New `graceSlashRoundTrip` / `graceSlashUnspecifiedOmitsAttribute` tests (`GraceCueApiTest.cpp`)
- [x] `make test`: all pass (4594 assertions in 366 test cases, plus the three examples)
- [x] `make test-api-roundtrip`: 153 passed, 0 failed (of 153 pinned; 1 newly unlocked)

## References

- Related: #312 (pre-existing grace-note collapse bug found while testing this fix)

### DIFF
--- a/src/include/mx/api/NoteData.h
+++ b/src/include/mx/api/NoteData.h
@@ -126,6 +126,9 @@ class NoteData
     // as 0 and is ignored on write), and cue notes -- including grace-cue
     // notes -- cannot carry <tie> (ties on them are silently dropped on write).
     bool isGrace;
+    // <grace>'s slash attribute. Only meaningful when isGrace is true.
+    Bool graceSlash;
+
     bool isCue;
 
     Notehead notehead;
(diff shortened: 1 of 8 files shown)


src/include/mx/api/NoteData.h
```diff
@@ -128,6 +128,10 @@ class NoteData
     bool isGrace;
     bool isCue;
 
+    // <grace>'s slash attribute (slashed grace notes, e.g. acciaccatura). Only meaningful
+    // when isGrace is true.
+    Bool graceSlash;
+
```
webern, line 131:
```suggestion
   bool isGrace;
    // <grace>'s slash attribute. Only meaningful when isGrace is true.
    Bool graceSlash;

      bool isCue;
```

PR #315, opened 2026-07-06
fix: round-trip the four accidental presence attributes

webern:
## Human Summary

TL;DR simple: add some missing attributes to the accidental element.

## Summary

`PitchData::isAccidentalParenthetical`/`Cautionary`/`Editorial`/`Bracketed` were already modeled and
read by `NoteReader`, but `NoteWriter` only ever wrote the accidental's value -- all four presence
attributes (`parentheses`, `cautionary`, `editorial`, `bracket`) were silently dropped on round-trip
(`attr:accidental@editorial`, `attr:accidental@parentheses`).

`NoteWriter` now sets each `core::Accidental` attribute to `"yes"` when its matching `PitchData` bool
is true, mirroring the reader's true-only-on-`"yes"` semantics exactly: a false bool omits the
attribute, matching the fact that the source was either absent or explicitly `"no"` (the reader
already can't distinguish those two cases) -- read/write stay symmetric.

Stacked on `#313` (grace note slash fix).

## Testing

- [x] New `AccidentalPresenceAttributesRoundTrip` test (`PitchDataTest.cpp`): all four attributes
      round-trip through XML and back
- [x] `make test`: all pass (4605 assertions in 367 test cases, plus the three examples)
- [x] `make test-api-roundtrip`: 157 passed, 0 failed (of 157 pinned; 4 newly unlocked -- the fix
      also covered two additional parenthesized-accidental fixtures beyond the two identified
      candidates)

## References

N/A

### DIFF
--- a/src/private/mx/impl/NoteWriter.cpp
+++ b/src/private/mx/impl/NoteWriter.cpp
@@ -112,6 +112,22 @@ core::Note NoteWriter::getNote(bool isStartOfChord) const
     {
         core::Accidental accidental;
         accidental.setValue(myConverter.convert(myNoteData.pitchData.accidental));
+        if (myNoteData.pitchData.isAccidentalParenthetical)
+        {
+            accidental.setParentheses(core::YesNo::yes());
+        }
+        if (myNoteData.pitchData.isAccidentalCautionary)
+        {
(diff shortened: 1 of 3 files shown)


webern:
@copilot rebase this over main and force push to the branch

PR #320, opened 2026-07-06
feat: unified spanner numbering with writer-side assignment

### LINKED ISSUE #297: Add API-level spanner identity for writer-side `number` assignment
### Problem

MusicXML uses `number` attributes on spanners such as `<slur>` and `<wedge>` to distinguish simultaneously open spanners of the same type.

For slurs, assigning these numbers correctly can depend on the final serialized MusicXML order, not just musical time. This is especially true when multiple voices are written using `<backup>`: a slur that has musically ended may still appear open to a streaming reader until the later voice containing its stop is serialized.

Currently, `mx::api` exposes independent start/stop objects with `numberLevel`, so callers must assign MusicXML `number` values before writing. But callers do not own the final serialization order.

### Proposed API

Add API-level spanner identity metadata to start/stop objects. This metadata would not be serialized directly. It would only tell the writer which start and stop belong to the same logical spanner.

Example shape:

```cpp
  struct CurveStart {
      CurveType curveType;
      int numberLevel;
      std::optional<std::string> spannerId; // API metadata only
      ...
  };

  struct CurveStop {
      CurveType curveType;
      int numberLevel;
      std::optional<std::string> spannerId; // API metadata only
      ...
  };
```

A similar identity field could be added to other spanner types such as wedges.

### Writer Behavior

When matching spanner identity metadata is present, the writer could assign or normalize MusicXML number values using the actual serialization order it controls.

Suggested behavior:

- Pair start/stop records by spanner type and spannerId.
- Assign legal MusicXML number values from 1..16.
- Track currently open spanners in serialized output order.
- Reuse the lowest available number when possible.
- Preserve existing explicit numberLevel behavior when no spannerId is present, or define a clear precedence rule.

### Why This Matters

A musical-time overlap calculation is not always enough.

Example:

- slur A: starts measure 1 beat 1, ends beat 3
- slur B: starts measure 1 beat 3.5, ends next measure beat 1
- slur C: starts next measure beat 2, ends next measure beat 3
- Musically, B and C do not overlap.
- But if voice 1 is serialized before voice 2 using <backup>, Slur C may be written before Slur B’s stop appears in the MusicXML stream.

In that serialized order, both slurs are simultaneously open to a streaming MusicXML reader, so they need different number values.

Because mx::api controls the final note serialization order, writer-side assignment based on API-level spanner identity would be more reliable than requiring callers to guess that order.

Here is a test case that illustrates the above scenario:

[slurs_overbars-ref.musicxml.txt](https://github.com/user-attachments/files/29668422/slurs_overbars-ref.musicxml.txt)

webern:
## Human Summary

Take @rpatters1&#39;s suggested design from #297 and attempts to implement it with a spanner identifying object and a sort of spanner resource pool.

## AI Summary

MusicXML pairs a spanner's start with its stop using a `number` attribute, and two spanners of the same type that are open at the same point in the serialized stream must carry different numbers. Which spanners are open at the same point depends on the order the writer emits notes (voice interleaving via `backup`), so a caller cannot reliably choose the number. This PR moves number assignment into the writer.

The design differs from the shape proposed in #297 (a second `spannerId` field beside `numberLevel`): two mutually exclusive fields side by side is the recurring defect that design principle #3 and #249 exist to eliminate. Instead the two are unified into one value type.

- New `SpannerNumber` value type (`SpannerNumber.h`/`SpannerNumber.cpp`, following the `TimeChoice.h`/`.cpp` declaration/implementation split) with exactly three states: unspecified (default, no number emitted for a lone spanner), explicit level 1..16 (emitted verbatim; what the reader produces), and identity (an author-only, never-serialized label; start/continue/stop sharing an id are the same logical spanner and the writer assigns the number). A contradictory state is unrepresentable by construction. Constructed directly (`SpannerNumber{3}`, `SpannerNumber{"my-id"}`), not via factory functions. Follows the `TimeChoice` "choice" pattern: nothing throws. An out-of-range level or an empty id collapses to unspecified at construction, and calling `level()`/`identity()` for the wrong kind returns a harmless default (`NUMBER_LEVEL_UNSPECIFIED`, or an empty string) instead of an exception.
- Breaking: `int numberLevel` is replaced by `SpannerNumber number` in `CurveStart`, `CurveContinue`, `CurveStop`, `WedgeStart`, `WedgeStop`, `SpannerStart`, and `SpannerStop`. Tuplets are untouched and keep the legacy `int` + `NUMBER_LEVEL_UNSPECIFIED`.
- New `SpannerNumberResolver` runs once per part before writing. It walks the part in true serialization order (measures, staves, voices ascending, notes in vector order with curve stops/continues/starts per note, directions in vector order) and assigns each identity spanner the lowest free number from a 1..16 pool per staff and per spanner class (slur, tied, wedge, octave-shift, bracket, and dashes pools are independent). While an explicit spanner is open its level is reserved, so explicit and identity spanners sharing a pool never collide. Pool exhaustion (more than 16 concurrently open) throws rather than emitting an illegal number.
- The wedge and octave-shift writers now emit numbers at all; previously they dropped them on write even when the reader had captured them, so explicit wedge/ottava numbers now survive a round trip.
- Readers only ever produce unspecified or explicit, keeping read-to-write fidelity for files that already carry numbers.

## Testing

- [x] `SpannerNumberTest`: three-state semantics, out-of-range/empty-input construction collapsing to unspecified, wrong-kind accessors returning harmless defaults (never throwing), equality across all kind combinations
- [x] `SpannerIdentityTest`: the #297 scenario — slurs disjoint in musical time that overlap in the stream once voice 1 serializes ahead of voice 2 get distinct numbers (1, 1-reused, 2); explicit + identity pool sharing; slur/tied pool independence; wedge number round-trip
- [x] `data/rpatters1/slurs_overbars.musicxml`: the file rpatters1 attached to #297 (a genuine Finale export, confirmed provenance in review), pinned in `roundtrip-baseline.txt` alongside his other contributed fixtures in `data/rpatters1/`
- [x] `make test`, `make test-api-roundtrip` (166/166 pinned), `make test-core-dev` (835 corert files), `make check-core-dev`, `make validate-cpp`, `make test-cpp-unit`, `make probe-cpp`
- [x] Unity build (`CMAKE_UNITY_BUILD=ON`, batch size 0) compiles
- [x] Rebased over main (through #333) and CI re-run on the rebased branch

## References

- Closes #297
- Related to #249 (one fact, one field: motivates unifying level and identity into a single type instead of adding a parallel field)


data/rpatters1/slurs_overbars.musicxml
```diff
@@ -0,0 +1,367 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!DOCTYPE score-partwise PUBLIC "-//Recordare//DTD MusicXML 4.0 Partwise//EN" "http://www.musicxml.org/dtds/partwise.dtd">
+<score-partwise version="4.0">
+  <identification>
+    <encoding>
+      <software>Finale v27.4 for Mac</software>
```
webern, line 6: Where did this file come from? Did we alter it from a file we had? Or is it completely synthetic? The reason I'm asking is that we have a directory named data/synthetic for files that the LLM created. And those files have a risk of being nonsensical. But I don't understand the provenance of this file so I can't understand if this is the best location and precedent for it.

Possibly the <software></software> tag would be a good place to describe how it was created?

rpatters1: It is an export from Finale 27.4. Do you need the musx that it came from?

webern: Oh, no that's perfect. I didn't realize we had pulled it in from the issue comments and thought maybe the LLM had invented it. 👍

webern:
Pushed a commit addressing the `pr feedback` TODOs:

- `SpannerNumber` moved out of `ApiCommon.h` into its own `SpannerNumber.h`/`SpannerNumber.cpp` (declaration + implementation split, matching `TimeChoice.h`/`.cpp`).
- `makeLevel`/`makeIdentity` factory functions replaced with constructors (`SpannerNumber{3}`, `SpannerNumber{"my-id"}`).
- No more throwing. Following the `TimeChoice` pattern: an out-of-range level or an empty id collapses to `unspecified` at construction, and calling `level()`/`identity()` for the wrong kind returns a harmless default (`NUMBER_LEVEL_UNSPECIFIED`, or an empty string) instead of throwing.

On the two `// TODO: ... Is this part of the spanner interface? Does it belong in ApiCommon or somewhere else?` comments on `Placement` and `MeasureNumbering`: I left those where they are. They predate this PR (this PR only inserted `SpannerNumber` above them, it didn't add them) and they're not spanner-specific — `Placement` is used across `CurveData.h`, `DefaultsData.h`, `DirectionData.h`, and `PositionData.h`, and `MeasureNumbering` by `MeasureData.h`. Moving them would be a separate, unscoped cleanup touching several unrelated headers, so I didn't fold it into this PR. Happy to do that as a follow-up if you'd like.

Also moved `slurs_overbars.musicxml`/`.features.xml` into `data/rpatters1/` per your last comment, updated `roundtrip-baseline.txt` and regenerated `data/corpus.xml`, and rebased over the `#333` merge to main (conflict was just the round-trip allow-list, resolved by keeping both new pins). CI is running on the new commit.

---

PR #321, opened 2026-07-06
feat(api): support the full time-signature model

### LINKED ISSUE #268: api: time signature improvements

webern:
## Human Summary

When @rpatters1 opened #268, my immediate thought was, oh no, we have to support all of the time signature model. My reasoning is that just adding the one case that is needed now will only last so long before a different semantic situation is encountered.

I wanted to preserve some sort of "simple" path, so I created a "choice" class where `TimeSignatureData` used to be. It can be either `simple` or `complex` where simple basically means `numerator/denominator` with or without the `C` or `Cut` symbol.

Anything *beyond* that is shuttled into `ComplexTimeSignature` where, I believe, we have modeled all of what MusicXML supports.

## !!Breaking!!

We use "breaking" somewhat loosely around here, since just about any `mx::api` change introduces some amount of change to the client code, but this one is bigger than the usual breakage. It requires edits anywhere time signatures are involved in client code. As such, I spent some extra time on it and I hope I have a stable design here.

## Summary

Redesigns the time-signature model in `mx::api` to cover the full MusicXML `<time>` element instead of a single beats/beat-type pair with a narrow symbol set.

- `TimeSignatureData` is now built around a shared `TimeFraction` leaf (`beats`/`beatType`); its `TimeSignatureSymbol` is narrowed to `unspecified`/`common`/`cut`.
- New `ComplexTimeSignature` (`ComplexTimeSignature.h`) quarantines everything unusual: composite/additive multi-fraction meters, the full symbol vocabulary (`singleNumber`, `note`, `dottedNote`), separators, interchangeable (dual) meters, and senza-misura.
- New `TimeChoice` (`TimeChoice.h`) is a `{simple, complex}` switch carrying the whole-`<time>` attributes (`isImplicit`, `display`). Building a `TimeChoice` from a `ComplexTimeSignature` that turns out to be simple-equivalent auto-collapses it back to `simple`, so there's always one canonical representation.
- `MeasureData::timeSignature` is now a `TimeChoice`; a new `MeasureData::staffTimeSignatures` (`std::map<int, TimeChoice>`) holds per-staff `<time number="N">` overrides.
- impl: `TimeReader` now translates every `<time>` in a measure (not just the first) into a `TimeChoice` + staff index; `PropertiesWriter::writeTime` regenerates all shapes and writes per-staff `number=`; `MeasureReader`/`MeasureWriter`/`Cursor` carry per-staff overrides forward measure-to-measure; `Converter` gains symbol/separator/relation enum tables.
- Four previously-failing corpus files now round-trip and are pinned in `roundtrip-baseline.txt` (`ly11d`/`ly11e` composite meters, `ly11f` symbols, `ly11h` senza-misura).

This is a breaking change to `mx::api`: `beats`/`beatType` move into `fraction`, `isImplicit`/`display` move from `TimeSignatureData` onto `TimeChoice`, and `TimeSignatureSymbol::singleNumber` moves to the complex-only `ComplexTimeSymbol`.

## Testing

- [x] `TimeSignatureApiTest.cpp`: new cases for default/auto-collapse/stays-complex/equality behavior, plus round-trip coverage for composite meters, senza-misura (with and without glyph), interchangeable pairs, widened symbol+separator, and per-staff overrides with carry-forward
- [x] `roundtrip-baseline.txt`: `ly11d`, `ly11e`, `ly11f`, `ly11h` unpinned (now pass the api round-trip)
- [x] Existing api/impl fixtures updated to the new `TimeChoice` construction (`ApiK007a/c`, `ApiK009bSlur`, `ApiK014aFermatas`, `ApiK015aLayout`, `ApiLy43e`, `ApiMuAccidentals1`, `ChordApiTest`)
- [ ] `timesigs_composite-ref.musicxml` which was submitted by @rpatters1 to cover his use case

## References

- Closes #268

### DIFF
--- a/src/include/mx/api/MeasureData.h
+++ b/src/include/mx/api/MeasureData.h
@@ -6,12 +6,9 @@
 
 #include "mx/api/ApiCommon.h"
 #include "mx/api/BarlineData.h"
-#include "mx/api/ClefData.h"
-#include "mx/api/DirectionData.h"
 #include "mx/api/KeyData.h"
 #include "mx/api/PartSymbolData.h"
 #include "mx/api/StaffData.h"
-#include "mx/api/TempoData.h"
 #include "mx/api/TimeChoice.h"
 #include "mx/api/TransposeData.h"
 
@@ -22,7 +19,7 @@ namespace mx
 {
 namespace api
 {
-/// This typedef help readability in a few places such as the layout map in ScoreData.
+// This typedef helps readability in a few places such as the layout map in ScoreData.
 using MeasureIndex = int;
 
 class MeasureData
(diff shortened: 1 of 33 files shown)


src/include/mx/api/MeasureData.h
```diff
@@ -31,8 +31,15 @@ class MeasureData
     // this is the notes and other music data
     std::vector<StaffData> staves;
 
+    // The unscoped <time> (applies to all staves). staffIndex is INDEX_UNSPECIFIED here.
     TimeSignatureData timeSignature;
 
+    // Normally empty: one entry per staff-scoped <time number="N">. Each entry's staffIndex is the
```
webern: I think this is fine and it works. It's just that the comment is a bit hard to understand. It should be something like this:

Only use this is if you need different time signatures on different staves.

Furthermore, we should not put a staff index on the time signature structure. Instead this should be a map or a vector of pairs naming the staff index and the time signature override that goes with it.

src/include/mx/api/TimeSignatureData.h
```diff
@@ -6,58 +6,261 @@
...
+#include <utility>
+#include <variant>
+#include <vector>
 
 namespace mx
 {
 namespace api
 {
+
+// -------- enums (unspecified-first, per mx::api convention) --------
```
webern: Don't make separator comments like this.

src/include/mx/api/TimeSignatureData.h
```diff
@@ -6,58 +6,261 @@
...
+
+// Maps to MusicXML time-symbol. 'normal' (the default fractional display) folds into 'unspecified':
+// an unspecified symbol writes no attribute, which the spec treats as normal.
 enum class TimeSignatureSymbol
 {
     unspecified,
     common,
     cut,
-    singleNumber
+    singleNumber,
```
webern: I don't think this should be a list of symbols but then have this one item that's not exactly a symbol but rather a marker that tells you (I guess?) look somewhere else for the single number that is the symbol?

src/include/mx/api/TimeSignatureData.h
```diff
@@ -6,58 +6,261 @@
...
+    slash,
+    space,
+    hyphen
+};
+
+// -------- public leaf: one beats/beat-type pair --------
+
+// beats/beatType are free-form strings by design: they may hold compound values ("3+2"),
+// decimals ("1 + 2.5"), or compound denominators ("8 + 16").
+struct TimeSignatureComponent
```
webern: I don't like this class because to me this looks like what a "normal" time signature should be. This is basically the one that I would want to interact with 99% of the time, but it's called a "component".

src/include/mx/api/TimeSignatureData.h
```diff
@@ -6,58 +6,261 @@
...
+};
+
+// -------- the main type --------
+
+// A single <time> element. The *signature shape* (simple / composite / senza-misura) is held in a
+// private variant and reached only through the accessors below; you never touch the variant, and the
+// accessors never throw. The orthogonal display/scoping attributes are plain public fields.
+//
+// Defaults to an implicit, simple 4/4.
 class TimeSignatureData
```
webern: I don't want this class buried under all this new stuff. I'd rather have all the new, complicated stuff in a separate file. Name TBD. TimeSignatureX where X is something like
Components, Support, Items, I don't know. Something to get this garbage interface out of the way of the simpler use cases.

src/include/mx/api/TimeSignatureData.h
```diff
@@ -6,58 +6,261 @@
...
+    }
+
+    // ---- discriminators (never throw) ----
+
+    bool isMetered() const
+    {
+        return std::holds_alternative<MeteredTimeSignature>(mySignature);
+    }
+
+    bool isSenzaMisura() const
```
webern: This interface is a disaster. I mean the whole thing is terrible. We need to try again.

I think we need to have TimeSignatureData for the simple use case, and something like ComplexTimeSignature for everything else.

Then maybe there's something like a TimeChoice class that switches between the two.

webern (commented):
Let's try again.

rpatters1:
I quite like the approach you have taken.

PR #326, opened 2026-07-07
fix: preserve staff-scoped `staff-layout number="N"`

### LINKED ISSUE #281: api: round-trip drops staff-layout
Split out of #219 (investigate api dropping elements marked support=full), one element per issue.

## What

`<staff-layout>` is dropped on api round-trip, affecting ~33 corpus files. `data/api.features.xml` audits it as `partial`.

To be determined (the point of this issue):
1. a genuine impl round-trip bug - we model `staff-layout` but drop it on write (impl fix), or
2. `api.features.xml` overstating support - it is really partial/none and the audit is wrong (audit correction).

## Repro

```
make dump-api-roundtrip && make classify-api-roundtrip
# build/api/classified.json: <staff-layout> among the drop signatures
```

## References

- Split from #219 · Part of #208

webern:
## Human Summary

TODO: human writes here

## Summary

Both the per-measure `<print>` and score-level `<defaults>` readers took only the first
`<staff-layout>`'s distance and dropped its `number` attribute entirely, folding it into the single
unscoped `SystemLayoutData::staffDistance`. For a multi-staff part (e.g. a 2-staff piano/organ)
whose only `staff-layout` is scoped to staff 2, this silently misattributed the distance to staff 1
(the schema default for a missing `number`) instead of just dropping the attribute -- a real,
if quiet, correctness bug, not only a fidelity loss.

Added `SystemLayoutData::staffDistances`, a `std::map<int, double>` keyed by zero-based staff
index, holding spacing for staves that need their own value; the existing singular
`staffDistance` keeps its meaning for the common unscoped case, and for a given staff a map entry
wins. The map shape follows the `MeasureData::staffTimeSignatures` precedent for staff-scoped
`number="N"` overrides -- key by the owning staff rather than stamping a label field on the items
-- which also makes two conflicting distances for the same staff unrepresentable. Both readers
(`ScoreReader::scanForSystemInfo`, `LayoutFunctions::addStaffLayout`) and both writers
(`MeasureWriter::writeSystemInfo`, `LayoutFunctions::addSystemMargins`) handle it symmetrically,
since `<defaults>` and per-measure `<print>` already share `SystemLayoutData`. Header comments are
written for the api user (what the fields mean and when to set them) per the design doctrine,
rather than describing reader/writer behavior.

## Testing

- [x] New `staffScopedStaffDistance` test (`PrintLayoutRoundTripTest.cpp`)
- [x] `make test`: all pass (4808 assertions in 391 test cases, plus the three examples)
- [x] `make test-api-roundtrip`: 164 passed, 0 failed (of 164 pinned; no regressions)

## References

- Closes #281
- Progresses #208, #213

### DIFF
--- a/src/include/mx/api/SystemLayoutData.h
+++ b/src/include/mx/api/SystemLayoutData.h
@@ -9,6 +9,7 @@
 #include "mx/api/LeftRight.h"
 #include "mx/api/PageLayoutData.h"
 
+#include <map>
 #include <string>
 #include <vector>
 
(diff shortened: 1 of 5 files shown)


src/include/mx/api/SystemLayoutData.h
```diff
@@ -16,6 +16,29 @@ namespace mx
 {
 namespace api
 {
+/// A <staff-layout> scoped to one staff of a multi-staff part (MusicXML's number attribute):
+/// the space between the bottom line of the previous staff and the top line of this one.
+/// staffIndex is zero-based; a source with an explicit number attribute always populates
```
webern: A source? Meaning a file?

Since the coding agent has been tasked with round trip success, it is writing API documentation from that vantage-point which would be confusing to a user. Instead this documentation should explain how someone using `mx::api` to build up a score should use the interface. And the interface should avoid adding confusing things that exist solely to represent what is not music-notation-semantics-meaningful in a round-trip corpus test file.

```text
/// staffIndex is zero-based; a source with an explicit number attribute always populates
```

PR #328, opened 2026-07-07
docs: confirm isTieStart+isTieStop already model the continuation tie

### LINKED ISSUE #291: support continuation tie in api
I'm wondering if we should create a similar class like `ContinuationTie` that models ties at the start of 2nd endings and the like.  Not for this PR but something to think about.

_Originally posted by @rpatters1 in https://github.com/webern/mx/pull/286#issuecomment-4869525987_

webern:
## Human Summary

Interesting. Instead of implementing a continuation tie for #291, the agent confirmed that it is not technically needed and wrote documentation to that effect.

## Summary

#291 asked whether `mx::api` needs a dedicated `ContinuationTie` class (by analogy with
`TieLetRing`, #286) to model "ties at the start of 2nd endings and the like."

Per the MusicXML spec (`core::TiedType`'s doc comment): "For other ties that are visually attached
to a single note, such as a tie leading into or out of a repeated section or coda, use two tied
elements on the same note, one start and one stop." `NoteData::isTieStart`/`isTieStop` are already
independent bools (unlike `TieLetRing`'s lone, unpaired mark, this is an ordinary tie on each side),
and `NoteWriter` already emits both `<tie>` and `<tied>` pairs when both are set
(`NoteWriter::addTie`, called for stop then start). So the exact scenario rpatters1 described
already round-trips with the existing model -- a new class would just be a second, redundant way to
say the same thing, which is what #249's "one fact, one field" principle is explicitly meant to
prevent.

Added a regression test proving it (a note with `isTieStop` and `isTieStart` both true survives
write -> read intact) and expanded the field's doc comment to state this directly, since the shape
isn't obvious from the field declarations alone -- closing the loop on the design question with a
concrete, working answer instead of new API surface.

## Testing

- [x] New `tieStopAndStartOnSameNote` test (`NoteDataTest.cpp`)
- [x] `make test`: all pass (4734 assertions in 379 test cases, plus the three examples)
- [x] `make test-api-roundtrip`: 158 passed, 0 failed (of 158 pinned; no regressions -- no api
  behavior changed)

## References

- Closes #291
- Related: #286 (introduced `TieLetRing`, the one tie shape that genuinely doesn't fit
  start/stop pairing), #249 (one fact, one field)

### DIFF
--- a/src/include/mx/api/NoteData.h
+++ b/src/include/mx/api/NoteData.h
@@ -113,6 +113,14 @@ class NoteData
 
     // One field, two encodings: on write these emit both <tie> (sound) and
     // <tied> (notation), so the two can never contradict each other.
+    //
+    // The two are independent, so both may be true on the same note: this is exactly how
+    // MusicXML represents "a tie leading into or out of a repeated section or coda" (spec
+    // wording) -- e.g. the last note of a first ending, which closes the tie arriving from
+    // the note before it (isTieStop) and opens a new one continued into the corresponding
+    // note of the second ending (isTieStart). No separate "continuation tie" concept is
+    // needed; see TieLetRing just below for the one tie shape that genuinely doesn't fit
+    // this start/stop pairing.
     bool isTieStart;
     bool isTieStop;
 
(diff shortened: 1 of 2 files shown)


webern:
Closing this PR as it is likely not correct.

PR #330, opened 2026-07-07
feat: support multi-note tremolos in mx::api

### LINKED ISSUE #329: feat: support multi-note tremolos in `mx::api`
## Summary

`mx::api` can represent single-note tremolos, but not measured multi-note tremolos. This prevents faithful authoring and round-tripping of a common MusicXML tremolo form.

## Details

MusicXML uses `<tremolo>` for both single-note and multi-note tremolos.

`mx::api` currently supports the single-note case, but it does not provide a way to represent the multi-note case written as paired tremolo ornaments on adjacent notes, e.g.:

```xml
<tremolo type="start">3</tremolo>
...
<tremolo type="stop">3</tremolo>
```
Because there is no public API representation for this notation shape, API clients cannot author it directly, and read/write round-trip cannot preserve it faithfully.

webern:
## Human Summary

I do not love this but I am going to merge it.

`MarkData` and `MarkType` interact in a terrible way. If your `MarkType` is something, like mordent, then `MarkData` might be used to carry some additional information relevant only to mordents. Rather than continue this practice I introduced a MarkDataChoice type for future additions to this pattern, the first of which is multi-note tremolos which, I guess, you can specify the number of tremolo marks for.

This is not a clear API and ultimately it stems from a lack of Rust enums, which are so so wonderful compared to C++. I guess the `std::variant` is C++'s answer to Rust enums, but it's ugly has exceptions and undefined behavior, so I'm hiding it behind these "choice" classes (this started with #321).

Ultimately, I hope this is clear enough for coding agents to understand and use correctly even if it is just, not a great interface.

## Summary

`mx::api` could represent single-note (glyph) tremolos but silently dropped the measured, two-note
form written as `<tremolo type="start">`/`<tremolo type="stop">` on adjacent notes. This adds it.

- `MarkType` gains `tremoloStart` and `tremoloStop`, marking the first and second note of a
  measured tremolo. The existing `tremoloSingleOne..Five` glyph marks are unchanged.
- `MarkData` gains a `choice` field (`MarkDataChoice`), a variant class (mirroring the
  `TimeChoice`/`ComplexTimeSignature` pattern) that carries payloads for mark types whose data
  doesn't fit the common fields. For `tremoloStart`/`tremoloStop` it holds `TremoloMarkData` with
  `tremoloMarks` (`std::optional<int>`), the slash count (0-8). The `tremoloSingle*` marks keep
  encoding their count in the enumerator itself, as before.
- MusicXML's `<tremolo>` has no `number` attribute (unlike slur/tie), so there is nothing to key a
  cross-note match on. Pairing the start on one note with the stop on a later note is left to the
  api consumer via document order, the same way `NoteData::isTieStart`/`isTieStop` already work.
- `OrnamentsFunctions.cpp` (read) and `NotationsWriter.cpp` (write) are updated symmetrically;
  `isMarkTremolo`/`isMarkOrnament` include the two new mark types. Unmeasured tremolos are still
  not representable and continue to be dropped as before.
- Per review, `MarkData.h` now documents which of its fields are common to all marks vs. specific
  to one mark type, and directs future mark-specific additions to a new `MarkDataChoice`
  alternative (`MarkDataChoice.h`) rather than a new ad hoc direct field.

## Testing

- [x] New tests in `NoteDataTest.cpp`: an in-memory round trip of two adjacent notes carrying
      `tremoloStart`/`tremoloStop` marks, and a read test against `data/synthetic/tremolo.3.0.xml`
- [x] `make test`: all pass (4821 assertions in 392 test cases, plus the three examples)
- [x] `make test-api-roundtrip`: 164 passed, 0 failed (of 164 pinned; no regressions)
- [x] `make discover-api-roundtrip`: no newly-attributable passes; the two corpus files with real
      start/stop tremolo content (`musuite/testTremolo.xml`, `lysuite/ly23e_Tuplets_Tremolo.xml`)
      still fail on unrelated pre-existing subset gaps (an `identification`/`rights` mismatch and a
      `time-modification` gap, respectively), so no `roundtrip-baseline.txt` entry was added

## References

- Closes #329

### DIFF
--- a/src/include/mx/api/MarkData.h
+++ b/src/include/mx/api/MarkData.h
@@ -7,6 +7,7 @@
 #include "mx/api/PositionData.h"
 #include "mx/api/PrintData.h"
 
+#include <optional>
 #include <string>
 
 namespace mx
@@ -87,6 +88,8 @@ enum class MarkType
     tremoloSingleThree, ///< A tremolo on a single note (a glyph, not a spanner) with 3 slashes
     tremoloSingleFour,  ///< A tremolo on a single note (a glyph, not a spanner) with 4 slashes
     tremoloSingleFive,  ///< A tremolo on a single note (a glyph, not a spanner) with 5 slashes
+    tremoloStart,       ///< The first note of a measured (two-note) tremolo; slash count is MarkData::tremoloMarks
+    tremoloStop,        ///< The second note of a measured (two-note) tremolo; slash count is MarkData::tremoloMarks
     otherOrnament,      ///< MusicXML's 'other-ornament' value
     unknownOrnament,    ///< Error state
 
@@ -252,6 +255,12 @@ struct MarkData
     Bool fingeringSubstitution;
     Bool fingeringAlternate;
 
+    // The tremolo mark count (MusicXML <tremolo> text value, 0-8 slashes). Only meaningful
+    // when markType is tremoloStart or tremoloStop; absent means "not specified" (the writer
+    // falls back to a default). The tremoloSingle* mark types encode their slash count in the
+    // enumerator itself and leave this absent.
+    std::optional<int> tremoloMarks;
+
     MarkData();
     MarkData(MarkType inMarkType);
     MarkData(Placement inPlacement, MarkType inMarkType);
@@ -271,6 +280,7 @@ MXAPI_EQUALS_MEMBER(mordentDeparture)
 MXAPI_EQUALS_MEMBER(hasMordentDeparture)
 MXAPI_EQUALS_MEMBER(fingeringSubstitution)
 MXAPI_EQUALS_MEMBER(fingeringAlternate)
+MXAPI_EQUALS_MEMBER(tremoloMarks)
 MXAPI_EQUALS_END;
 MXAPI_NOT_EQUALS_AND_VECTORS(MarkData);
 } // namespace api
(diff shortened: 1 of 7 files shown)


src/include/mx/api/MarkData.h
```diff
@@ -252,6 +255,12 @@ struct MarkData
     Bool fingeringSubstitution;
     Bool fingeringAlternate;
 
+    // The tremolo mark count (MusicXML <tremolo> text value, 0-8 slashes). Only meaningful
+    // when markType is tremoloStart or tremoloStop; absent means "not specified" (the writer
+    // falls back to a default). The tremoloSingle* mark types encode their slash count in the
+    // enumerator itself and leave this absent.
+    std::optional<int> tremoloMarks;
```
webern: The MarkData class is really quite a mess. It has fields that are only relevant when the mark is of the associated type that they are relevant to.

Here's what I think we should do. We should create a MarkDataChoice case similar to how we recently handled time signatures (follow that pattern of variant use without throwing or UB)

Any new items that need to go on MarkData but only apply to certain marks should go in the choice field. The choice field SHOULD align with the selected mark type but it is not enforced.

We document the fields of MarkData that are (more or less) common to all marks, and separate them from existing fields (with documentation) from fields that only apply to certain marks, and we document that new such cases should be added to the markdatachoice class instead of being added as direct fields.

PR #333, opened 2026-07-11
feat: add score-part groups, measure-numbering multi-rest/system attrs, and swing to mx::api

### LINKED ISSUE #102: MeasureNumbering might be dead code
See if anything is actually implementing MeasureNumbering. If not, remove it.

The enum is in `ApiCommon.h`

```C++
        enum class MeasureNumbering
        {
            unspecified,
            none,
            measure,
            system
        };
```

It is used in `MeasureData` and `DefaultsData` but unclear if anything reads or writes it to/from MusicXML.

webern:
## Human Summary

#268 introduced a test file that had a few more unsupported goodies in it. I want to support the file in full so this PR/commit adds:
- score-part group
- measure-numbering attributes
- and sound swing support

## Summary

`data/rpatters1/timesigs_composite-ref.musicxml` (the untrimmed original of the fixture that
motivated the `TimeSignatureData` redesign in #268) carried three things `mx::api` didn't model:
`<score-part>/<group>`, the `<measure-numbering>` `multiple-rest-always`/`multiple-rest-range`/
`system` attributes, and `<sound>/<swing>`. A trimmed `-modified` variant was pinned instead so
the composite time-signature round trip could land in isolation. This PR adds the three missing
features so the original file passes too:

- `PartData::groups` (`std::vector<std::string>`) models `<score-part>`'s repeatable `<group>`
  element, wired through `PartReader`/`PartWriter`.
- `MeasureData::measureNumberingMultipleRestAlways`/`measureNumberingMultipleRestRange` (`Bool`)
  and `measureNumberingSystemRelation` (new `SystemRelation` enum) model the `<measure-numbering>`
  attributes beyond its text value, wired through `MeasureReader`/`MeasureWriter` and a new
  `Converter::convertSystemRelation` bridge.
- `SoundData::swing` (new `SwingData`/`SwingNoteType`) models `<sound>/<swing>` (straight vs.
  ratio, note type, style), wired through `SoundFunctions.cpp`. `SoundData.h`'s "intentionally not
  modeled" comment is updated to drop `<swing>` from the list.

`data/rpatters1/timesigs_composite-ref.musicxml`'s `music-font` `font-family` is comma-only
(`"Maestro,engraved"`), which doesn't match mx's normalized comma-space output. Rather than edit
the fixture, the corert `normalizeForComparison` pipeline now re-spaces any comma-separated-text
attribute (currently just `font-family`, the schema's only user of that type) to `"item, item"` on
both sides of the round-trip comparison (new `CommaSeparatedFields.h`, mirroring the existing
`DecimalFields.h`/`stripZerosFromDecimalFields` pattern) — so the fixture stays byte-identical to
its original form.

`timesigs_composite-ref.musicxml` is now pinned in `roundtrip-baseline.txt` alongside the existing
`-modified` variant.

Also removes `DefaultsData::measureNumbering` (`src/include/mx/api/DefaultsData.h`), which was
dead code: MusicXML's `<defaults>` element has no `measure-numbering` child in the schema, so the
field could never be read from or written to a file. The real, working feature is
`MeasureData::measureNumbering` above, which is measure-scoped to match how `<measure-numbering>`
actually appears in MusicXML (inside a per-measure `<print>`).

## Testing

- [x] New unit tests: `ScorePartGroupApiTest.cpp` (group round trip), `SoundApiTest.cpp` swing
      cases, `MeasureDataTest.cpp` `measureNumberingAttributesRoundTrip`
- [x] `make test-api-roundtrip`: 165 passed (of 165 pinned), including the newly-added
      `rpatters1/timesigs_composite-ref.musicxml`
- [x] `make test`: all tests passed (4831 assertions in 396 test cases)
- [x] `make check-core-dev`: fmt-check and warning-free build both pass
- [x] `make test-core-dev`: all corert tests passed (835 test cases)

## References

- Closes #102
- The fixture originates from #268 (time signature improvements)

### DIFF
--- a/data/rpatters1/timesigs_composite-ref.musicxml
+++ b/data/rpatters1/timesigs_composite-ref.musicxml
@@ -51,7 +51,7 @@
       <distance type="hyphen">60</distance>
       <distance type="beam">7.5</distance>
     </appearance>
-    <music-font font-family="Maestro,engraved" font-size="20.5"/>
+    <music-font font-family="Maestro, engraved" font-size="20.5"/>
     <word-font font-family="Times New Roman" font-size="10.25"/>
   </defaults>
   <part-list>
--- a/data/rpatters1/timesigs_composite-ref.musicxml
+++ b/data/rpatters1/timesigs_composite-ref.musicxml
@@ -51,7 +51,7 @@
       <distance type="hyphen">60</distance>
       <distance type="beam">7.5</distance>
     </appearance>
-    <music-font font-family="Maestro, engraved" font-size="20.5"/>
+    <music-font font-family="Maestro,engraved" font-size="20.5"/>
     <word-font font-family="Times New Roman" font-size="10.25"/>
   </defaults>
   <part-list>
--- a/src/include/mx/api/SoundData.h
+++ b/src/include/mx/api/SoundData.h
@@ -6,17 +6,52 @@
 
 #include "mx/api/ApiCommon.h"
 
+#include <optional>
 #include <string>
 
 namespace mx
 {
 namespace api
 {
+// The swing-type-value type: which note type (eighth or 16th) a SwingData ratio applies to.
+// unspecified means the source did not state it, which MusicXML treats as "eighth".
+enum class SwingNoteType
+{
+    unspecified,
+    eighth,
+    dur16th
+};
+
+// MusicXML Documentation: The swing element specifies whether or not to use swing playback, where
+// consecutive on-beat/off-beat eighth or 16th notes are played with unequal nominal durations.
+struct SwingData
+{
+    // true: <straight/>, no swing, consecutive notes have equal durations. false: a swing ratio
+    // is in effect, described by ratioFirst/ratioSecond/noteType below.
+    bool isStraight;
+
+    // The swing ratio's two terms (e.g. 2 and 1 for a 2:1 ratio). Only meaningful when
+    // isStraight is false.
+    int ratioFirst;
+    int ratioSecond;
+
+    // Which note type the ratio applies to. Only meaningful when isStraight is false.
+    SwingNoteType noteType;
+
+    // A free-text description of the swing style (e.g. "Swing"). Empty means unspecified.
+    std::string style;
+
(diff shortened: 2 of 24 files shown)


data/rpatters1/timesigs_composite-ref.musicxml
```diff
@@ -51,7 +51,7 @@
       <distance type="hyphen">60</distance>
       <distance type="beam">7.5</distance>
     </appearance>
-    <music-font font-family="Maestro,engraved" font-size="20.5"/>
+    <music-font font-family="Maestro, engraved" font-size="20.5"/>
```
webern: No. This should be fixed for any affected file (that has a no-spaces csv pattern in this element) by correcting the "expected" side of the test comparison programmatically.

webern: Fixed in cc99979: the fixture is reverted to byte-identical with `main` (`"Maestro,engraved"`, no space). Instead, the corert `normalizeForComparison` pipeline now re-spaces any comma-separated-text attribute (currently just `font-family`, the schema's only user of that type) to `"item, item"` on both sides of the comparison — new `CommaSeparatedFields.h` + `normalizeCommaSeparatedFields`, mirroring the existing `stripZerosFromDecimalFields`/`DecimalFields.h` pattern. This covers any file with an unspaced font-family list, not just this one.

---

src/include/mx/api/SoundData.h
```diff
@@ -6,17 +6,52 @@
...
+
 // MusicXML Documentation: The sound element contains general playback parameters. They can stand
 // alone within a part/measure, or be a component element within a direction.
 //
-// mx::api models the commonly-used scalar attributes of <sound>. The nested child elements
-// (<midi-instrument>, <midi-device>, <play>, <swing>, <offset>) are intentionally not modeled.
+// mx::api models the commonly-used scalar attributes of <sound>, plus <swing>. The remaining
+// nested child elements (<midi-instrument>, <midi-device>, <play>, <offset>) are intentionally
+// not modeled.
 //
```
webern: Remove the statement about what is not modeled. The lack of modeling here is not a conscious decision currently from the owner and saying what's not there really isn't necessary.

PR #348, opened 2026-07-12
feat: round-trip the midi-device port and id attributes

### LINKED ISSUE #347: api: midi-device port and id attributes are dropped; attribute-only midi-device vanishes
## What

`MidiData::device` models only the `<midi-device>` element's text content. The `port` and `id`
attributes are dropped on read, and a midi-device whose text is empty (the common
MuseScore/Finale form `<midi-device id="P1-I1" port="1"></midi-device>`) is dropped entirely on
write, because the writer's emit condition is `device.size() > 0`.

Verified on `custom/musescore-slur-start-stop.musicxml`:

- expected: `<midi-device id="P1-I1" port="1" />`
- actual: (element missing)

## Impact (classifier signature `drop:midi-device`)

41 candidate files, 3 of them distance-1 (`custom/musescore-slur-start-stop.musicxml`,
`musuite/testChord.xml`, `musuite/testPiano.xml`).

## Fix shape

Add `MidiData::devicePort` / `MidiData::deviceId` (`std::optional`), read them from
`core::MIDIDevice`, and emit the element when any of text/port/id is present.

## References

- Part of #208

webern:
## Human Summary

TODO: human writes here

## Summary

The midi-device port and id attributes were dropped on read, and the common attribute-only form
(`<midi-device id="P1-I1" port="1"></midi-device>`, emitted by MuseScore and Finale) vanished on
write because the emit condition required non-empty device text. This exposes the port, keeps the
id round-tripping, and emits the element whenever device text or a port is present.

Issue #347 suggested storing the id as a `deviceId` field. Instead the id is modeled by
containment. The midi-device id is a back-reference to the score-instrument the device belongs to,
and mx::api already models that instrument as the part's single `InstrumentData` — so a stored id
would just duplicate the instrument's `uniqueId` and force callers to keep the two in sync.

- The writer derives the midi-device id from the part instrument's `uniqueId`, the same id the
  sibling midi-instrument already writes. There is no separate id for a caller to manage.
- `MidiData::writeDeviceId` (unspecified/yes/no, default unspecified) records only whether the
  source spelled the link out, so an id-less midi-device stays id-less and an id-bearing one is
  preserved. Same fidelity-knob convention as `ClefData::writeStaffNumber`; authors can ignore it.
- `MidiData::devicePort` carries the 1–16 port.

While here, `InstrumentData` is documented to explain a real limitation: mx::api models one
instrument per part and does not support instrument changes mid-piece. MusicXML allows several
instruments per part (notably drum kits, where each sound is its own instrument); only the first is
brought in. That gap is left for a future change, since there is no clean way yet to expose it
without complicating the common single-instrument case.

## Testing

- [x] New `midiDeviceRoundTrip.portSurvivesWriteAndRead` and `writeDeviceIdSurvivesWriteAndRead`
- [x] Full api/impl suite passes (4983 assertions in 427 test cases)
- [x] Round-trip baseline 224 -> 227 (musescore-slur-start-stop, testChord, testPiano); the id-less
  midi-device in ksuite/k008a_Beaming.xml continues to pass

## References

- Closes #347
- Part of #208

### DIFF
--- a/src/include/mx/api/PartData.h
+++ b/src/include/mx/api/PartData.h
@@ -28,15 +28,26 @@ struct MidiData
 {
     std::string virtualLibrary;
     std::string virtualName;
+
+    // The name of the playback device or sound bank for this instrument — for example the name of a
+    // MIDI output device, or a sound or bank label. This is often left empty.
     std::string device;
 
-    // The midi-device port attribute: a number from 1 to 16 for the unofficial MIDI 1.0
-    // port (cable) meta event.
+    // An optional port (or cable) number, from 1 to 16, that says which output the device uses. It
+    // has no value when the file doesn't specify one, and most files leave it unset — you only need
+    // it when you are aiming at a particular MIDI port. You don't have to attach the device to this
+    // part's instrument yourself; the library keeps them together automatically when the music is
+    // written out.
     std::optional<int> devicePort;
 
-    // The midi-device element's id attribute, which refers to the score-instrument the
-    // device assignment belongs to.
-    std::optional<std::string> deviceId;
+    // Most users can ignore this; leave it unspecified. It only controls whether the saved file
+    // spells out the little link that ties this part's playback device to its instrument. unspecified
+    // (the default) does the sensible thing on its own: it leaves that link out, because each part
+    // here has a single instrument and so the link isn't needed. yes/no force it to be written or left
+    // out. You never have to set this when you are creating music — when the library reads a file that
+    // did spell the link out, it sets this for you, so that saving the file brings it back the same way
+    // it came in.
+    Bool writeDeviceId;
 
     std::string name;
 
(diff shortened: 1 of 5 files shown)


src/include/mx/api/PartData.h
```diff
@@ -29,6 +29,15 @@ struct MidiData
...
     std::string virtualName;
     std::string device;
+
+    // The midi-device port attribute: a number from 1 to 16 for the unofficial MIDI 1.0
+    // port (cable) meta event.
+    std::optional<int> devicePort;
+
+    // The midi-device element's id attribute, which refers to the score-instrument the
+    // device assignment belongs to.
+    std::optional<std::string> deviceId;
```
webern: Hmm first of all there's something about saying "The midi-device attribute" and in the next one saying "The midi-device element's attribute" that throws me off. It makes me ask, wait, are these on the same element? And if so why are these worded differently.

Secondly, I don't like things that point somewhere else, like `deviceId` is doing here, when it could simply be a dangling pointer and forces bookkeeping on the client to make sure they mark the same `deviceId` here as they do somewhere else. Is there another way to model it?

PR #351, opened 2026-07-13
fix: scope spanner number pools to the part, not the staff

### LINKED ISSUE #350: SpannerNumberResolver pools numbers per staff, but `number-level` scope is the part
### Summary

`SpannerNumberResolver` (#320) pools spanner numbers per **staff** per class (`occupied[staffIndex][number]`), but MusicXML's `number-level` type scopes concurrency to the **part**: numbers distinguish "concurrent objects of the same type when the objects overlap in MusicXML document order." Its own piano example says a cross-staff slur "will need a separate number from the mid-measure slurs" even though they are on different staves.

### Repro

Two-staff part, two identity slurs: staff 2 spanning m1 -> m2, staff 1 inside m2. Serialized order:

```
m1 staff 2:  <slur type="start" number="1"/>   (still open when the next start emits)
m2 staff 1:  <slur type="start" number="1"/>
m2 staff 1:  <slur type="stop"  number="1"/>
m2 staff 2:  <slur type="stop"  number="1"/>
```

The second slur must get a different number. Readers pair slur endpoints by number within the part, so this duplicate typically pairs the m1 start with the m2 staff-1 stop ‚Äî a phantom cross-staff slur ‚Äî and drops the rest.

### Suggested fix

Drop the staff dimension from `occupied` and pool per part + class only. The cross-staff "reserve in every staff pool touched" logic then becomes unnecessary.

### File containing this example.

(Note: this file was created by my new Finale exporter and MX because Finale's current exporter exports every staff as a different part. That means this file exhibits the bug.)

[numberlev_2staffpart.musicxml.txt](https://github.com/user-attachments/files/29947903/numberlev_2staffpart.musicxml.txt)

rpatters1:
## Summary

Fixes #350. `SpannerNumberResolver` pooled spanner numbers per **staff**, but MusicXML's `number-level` type scopes concurrency to the **part** in document order ‚Äî its own piano example requires a cross-staff slur to take a different number from staff-local slurs it overlaps. Two identity slurs overlapping in document order on different staves of the same part therefore both received `number="1"`, and readers that pair endpoints by number within the part (e.g. MuseScore) mispaired them.

## Fix

Drop the staff dimension from the resolver: events no longer carry a staff index, and explicit reservations and identity groups live in one pool per part + spanner class. The cross-staff "reserve in every staff pool touched" logic disappears rather than needing repair.

## Fixed File

This file is the same one I posted in the issue, but it now has non-conflicting number levels.

[numberlev_2staffpart_fixed.musicxml.txt](https://github.com/user-attachments/files/29948432/numberlev_2staffpart_fixed.musicxml.txt)

## Testing

- [x] New: `overlappingSlursOnDifferentStavesGetDistinctNumbers` ‚Äî barline-crossing slurs on two staves of one part take numbers 1 and 2 (fails against the unfixed resolver: both came back 1)
- [x] New: `sameMeasureSlursOnDifferentStavesReuseTheNumber` ‚Äî the spec-blessed reuse case still gets 1/1, guarding against a naive always-increment fix
- [x] `make test` (425 cases), `make test-api-roundtrip` (224/224 pinned), `make fmt`
- [x] Unity build (`CMAKE_UNITY_BUILD=ON`, batch size 0) compiles
- [x] Verified end-to-end in a denigma export: the previously colliding two-staff slur fixture now emits distinct numbers

### DIFF
--- a/src/private/mx/impl/SpannerNumberResolver.cpp
+++ b/src/private/mx/impl/SpannerNumberResolver.cpp
@@ -9,7 +9,6 @@
 #include <cstddef>
 #include <limits>
 #include <map>
-#include <set>
 #include <stdexcept>
 #include <string>
 #include <vector>
@@ -36,7 +35,6 @@ enum class SpannerNumberClass
 struct SpannerNumberEvent
 {
     int position;
-    int staffIndex;
     const void *object;
     const api::SpannerNumber *number;
     bool opens;  // a start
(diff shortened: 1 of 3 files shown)


webern (approved):
Oops, thank you for fixing my mistake. Nice tests, too.

PR #353, opened 2026-07-14
fix: support optional member detection with MSVC

### LINKED ISSUE #352: Windows build fails with compiler errors
When I attempt to build on Windows, I get the following error messages.

Image

These are real compiler errors, not warnings. I'm not sure why the macOS builds work. I am using VS2022 on Windows 11.

rpatters1:
## Summary

Fixes the mx::api build with MSVC by making optional-member setter detection two-stage. Resolves #352.

The helper macros now confirm that a generated core type exposes the field accessor before evaluating setter expressions that depend on the field’s optional value type. This prevents MSVC from diagnosing absent attributes such as font decoration and stop-length fields on types that do not support them.

## Validation

- Configured and built `mx` with CMake and Visual Studio 2022/MSVC.
- `mx.lib` links successfully.
- `git diff --check` passes.

## Notes

- There are still a number of pedantic lint warnings with `/W4`, so users should use `/W3` for MSVC builds.
- There is a potential name conflict in `FontData.h` with the `small` macro in the WinAPI header `rpcndr.h`. This PR does not address that issue.

### DIFF
--- a/src/private/mx/utility/OptionalMembers.h
+++ b/src/private/mx/utility/OptionalMembers.h
@@ -101,22 +101,22 @@
                                                                                                                        \
     template <class T> bool lookForAndSetHas##fieldNameCapitalized(bool has, T *const el)                              \
     {                                                                                                                  \
-        if constexpr (requires { el->setterName(std::nullopt); })                                                      \
+        if constexpr (requires { el->fieldName(); })                                                                   \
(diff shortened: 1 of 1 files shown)


webern (approved):
Thank you. I'll get Windows added to CI soon.

Issue #324, opened 2026-07-07
api: many direction-type subtypes are unmodeled stubs (percussion, metronome extras, harp-pedals, scordatura, ...)


webern:
> This is an AI written issue:

Split out of #278/#280 while investigating why the round-trip classifier reports `drop:voice` /
`drop:staff` / `drop:direction` / `drop:direction-type` on ~60 corpus files.

## Finding

Nearly all of those drops trace to one cause: `DirectionReader` already dispatches every
`direction-type` choice, but a long list of subtypes are no-op stubs (`parseHarpPedals`,
`parseDamp`, `parseDampAll`, `parseEyeglasses`, `parseStringMute`, `parseScordatura`, `parseImage`,
`parsePrincipalVoice`, `parseAccordionRegistration`, `parsePercussion`, `parseOtherDirection` --
all `MX_UNUSED(directionType);`). When a `<direction>` contains only one of these, the resulting
`DirectionData` ends up empty and is correctly not written back -- MusicXML requires at least one
`direction-type` child, so there is no schema-valid way to keep the `<direction>` (and its
`<voice>`/`<staff>`) without modeling its actual content.

`data/api.features.xml` already correctly audits each of these specific subtypes as `support="none"`
(only the parent `direction`/`direction-type` container is `full`, correctly -- the container
mechanism itself works). So this is not an audit-correction; it's a real, if long, worklist of
unmodeled features.

## Worklist (files unlocked once implemented; each also needs `add:attributes`/`add:divisions`
handled for the synthetic single-feature fixtures, so implementing one of these may not by itself
flip a file to PASS)

- `percussion` (beater/effect/glass/membrane/metal/other-percussion/pitched/stick/stick-location/
  stick-material/stick-type/timpani/wood) -- 24 files
- `metronome-note`/`metronome-relation`/`metronome-type`/`metronome-arrows` (metric modulation) -- 8 files each
- `accordion-registration` -- 3 files
- `damp`, `damp-all`, `effect`, `eyeglasses`, `harp-pedals`, `image`, `other-direction`,
  `principal-voice`, `string-mute` -- 2 files each
- `scordatura`, `staff-divide`, `symbol` -- 1 file each

Also found while investigating: `DirectionReader::parsePedal` only handles `PedalType::start`/`stop`;
`sostenuto`/`change`/`continue`/`discontinue`/`resume` are silently dropped
(`lysuite/ly33a_Spanners.xml`). Modeling these needs a shape decision (a third "kind" on the shared
[...]

PR #365, opened 2026-07-19
feat: round-trip octave-shift size

webern:
## Human Summary

Apparently, emitting "8" for a single octave shift is optional. Here we add a round-trip fidelity knob that can actually be ignored when using the API.

## Summary

octave-shift's `size` (8 or 15) is derived from `OttavaType`, so `emitOttavaStart` force-wrote `size="8"` -- the spec default -- on every 8va/8vb line. `OttavaStart::writeDefaultSize` (a plain-bool principle-7 fidelity knob, in the mold of `ClefData::writeStaffNumber`) records whether the source spelled out that redundant `size="8"`; the writer now omits it by default while always emitting the load-bearing, non-default `size="15"`. The reader sets the flag only for an 8va/8vb line that actually carried the attribute, so a file that did write `size="8"` still round-trips it.

The knob governs only the redundant default: a plain `bool` (not the ternary `Bool`) because the only real degree of freedom is whether to show `size="8"` on an 8-line. There is deliberately no way to suppress the load-bearing `size="15"` on a 15-line -- that state would be corrupting, so it is made unrepresentable.

Surfaced while hill-climbing `lysuite/ly31a_Directions.xml` to a full round-trip PASS.

## Testing

- [x] Full unit suite passes (5130 assertions in 451 test cases)
- [x] api round-trip discovery: no regressions; `lysuite/ly33d_Spanners_OctaveShifts.xml` moves toward PASS from the size fix. The pins land in the stacked round-trip PR (#366).

## References

- Progresses #324
- First of a three-PR stack: octave-shift size (this) &rarr; direction offset (#367) &rarr; round-trip canonicalizers + pins (#366)

### DIFF
--- a/src/include/mx/api/OttavaData.h
+++ b/src/include/mx/api/OttavaData.h
@@ -28,7 +28,14 @@ class OttavaStart
     SpannerStart spannerStart;
     OttavaType ottavaType;
 
-    OttavaStart() : spannerStart{}, ottavaType{OttavaType::unspecified}
+    // Octave-shift size fidelity knob. The size value (8 or 15) follows from ottavaType, so a
+    // 15ma/15mb line always writes size="15" while an 8va/8vb line omits the redundant, spec-
+    // default size="8". When true, the writer also emits that redundant size="8"; the reader sets
+    // it when the source spelled the attribute out. Leave false (the default) when authoring. It
+    // has no effect on a 15ma/15mb line, whose load-bearing size is always written.
+    bool writeDefaultSize;
+
+    OttavaStart() : spannerStart{}, ottavaType{OttavaType::unspecified}, writeDefaultSize{false}
     {
     }
 };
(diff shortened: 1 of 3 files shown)


src/include/mx/api/OttavaData.h
```diff
@@ -28,7 +28,14 @@ class OttavaStart
     SpannerStart spannerStart;
     OttavaType ottavaType;
 
-    OttavaStart() : spannerStart{}, ottavaType{OttavaType::unspecified}
+    // Octave-shift size fidelity knob. The size value (8 or 15) follows from ottavaType, so a
+    // 15ma/15mb line always writes size="15" while an 8va/8vb line omits the redundant, spec-
+    // default size="8". When true, the writer also emits that redundant size="8"; the reader sets
+    // it when the source spelled the attribute out. Leave false (the default) when authoring. It
+    // has no effect on a 15ma/15mb line, whose load-bearing size is always written.
```
webern:
```suggestion
    // has no effect on a 15ma/15mb line, whose size is always written.
```

PR #370, opened 2026-07-28
feat: multirest roundtrip and `useSymbols` in mx::api

rpatters1:
## Human Summary

This completes basic multimeasure-rest support in `mx::api`, including whether short rests use individual rest symbols or a single multimeasure-rest shape.

  ## Summary

`MeasureData::multiMeasureRest` already represented the number of measures in a multimeasure rest, and `MeasureReader` populated it from `<multiple-rest>`. The writer did not consume the field, so the element was dropped when writing a `ScoreData`.

This adds the missing write path:

- `PropertiesWriter` creates `<measure-style><multiple-rest>`.
- `MeasureWriter` emits it when `multiMeasureRest` is greater than zero.
- Nonpositive values are omitted because MusicXML requires a positive integer.

The change also adds `Bool multiMeasureRestUseSymbols` to `MeasureData` for the MusicXML `use-symbols` attribute:

- `yes` and `no` are read and written explicitly.
- `unspecified` omits the attribute.
- The field is ignored when the multimeasure-rest count is nonpositive.

The change is additive and non-breaking. Existing authored scores behave as before unless they set a positive `multiMeasureRest` value.

Seven corpus files now survive the API round trip and are pinned in the regression baseline:

- `lysuite/ly02a_Rests_Durations.xml`
- `lysuite/ly02c_Rests_MultiMeasureRests.xml`
- `lysuite/ly02d_Rests_Multimeasure_TimeSignatures.xml`
- `lysuite/ly03d_Rhythm_DottedDurations_Factors.xml`
- `musuite/testMultiMeasureRest1.xml`
- `musuite/testMultiMeasureRest2.xml`
- `musuite/testMultiMeasureRest3.xml`

## Testing

- [x] `multiMeasureRestRoundTrip_MeasureData` verifies the element, count, `use-symbols="yes"`, and read/write round trip
- [x] `nonPositiveMultiMeasureRestOmitsElement_MeasureData` verifies that invalid counts are omitted even when `useSymbols` is set
- [x] Full API suite passes: 5201 assertions in 460 test cases
- [x] API round-trip regression: 295 passed, 0 failed
- [x] API round-trip discovery: seven newly passing files and no regressions
- [x] Changed files formatted and `git diff --check` passes

### DIFF
--- a/src/include/mx/api/MeasureData.h
+++ b/src/include/mx/api/MeasureData.h
@@ -70,6 +70,10 @@ class MeasureData
     // by this.
     int multiMeasureRest;
 
+    // When yes, use 1-, 2-, and 4-measure rest symbols instead of a single multimeasure-rest shape.
+    // Meaningful only when multiMeasureRest is greater than zero.
+    Bool multiMeasureRestUseSymbols;
+
     // The implicit attribute is set to "yes" for measures where the measure number should never
     // appear, such as pickup measures and the last half of mid-measure repeats. The value is "no"
     // if not specified.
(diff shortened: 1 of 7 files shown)


webern (approved):
Nice

PR #371, opened 2026-07-28
feat: extend octave-shift support in mx::api

rpatters1:
## Human Summary

Fills in two gaps in ottava support: octave-shift start formatting was read but not written, and three-octave 22ma/22mb lines could not be represented in mx::api.

## Summary

Extends the existing ottava model and writer in two areas:

- `DirectionWriter::emitOttavaStart` now writes the start's `PositionData` and `PrintData`, including default/relative positioning, font, and color. The reader already populated
  these fields, making this a writer-only omission.
- `OttavaType` gains `o22ma` and `o22mb` for three-octave shifts.
- The reader maps MusicXML `size="22"` and `type="down"` / `type="up"` to `o22ma` / `o22mb`, following MusicXML's reversed written-pitch direction.
- The writer emits `size="22"` with the corresponding MusicXML direction.
- Existing 8va/8vb default-size behavior is unchanged, as is `OttavaStop::size`.

The public API change is additive.

## Testing

- [x] New API XML round-trip coverage for `o22ma` and `o22mb`
- [x] New focused reader tests for both 22-line directions
- [x] Writer tests cover size, direction, position, font, and color
- [x] Full unit suite passes: 5,224 assertions in 463 test cases
- [x] API round-trip regression passes: 295/295 pinned files
- [x] Full API round-trip discovery shows no regressions
- [x] Formatting and diff checks pass

### DIFF
--- a/src/include/mx/api/OttavaData.h
+++ b/src/include/mx/api/OttavaData.h
@@ -19,7 +19,9 @@ enum class OttavaType
     o8va,  // octave up (writes octave-shift type="down")
     o8vb,  // octave down (writes octave-shift type="up")
     o15ma, // 2 octaves up (writes octave-shift type="down")
-    o15mb  // 2 octaves down (writes octave-shift type="up")
+    o15mb, // 2 octaves down (writes octave-shift type="up")
+    o22ma, // 3 octaves up (writes octave-shift type="down")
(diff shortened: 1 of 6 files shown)


webern (approved):
Oops. That seems better.

PR #373, opened 2026-07-29
feat: add measure text and measure-numbering staff to mx::api

rpatters1:
## Human Summary

Add fields for `<measure text="...">` and ` <measure-numbering staff="N">` to `MeasureData`. With the number of measure-number-related fields up the 3 in `MeasureData` it may make sense to refactor them into dedicated `MeasureNumberData` class. This would be a breaking change, so I did not include that in this PR. I could rework it that way, though, if it makes sense to do so.

## Summary

Two `<measure>`-scoped attributes were being dropped on import and never written on export. Both
are now exposed on `MeasureData` and wired symmetrically through `mx::impl`.

`displayedNumber` (`std::optional<std::string>`) carries `<measure text="...">`: the measure
number as printed, when that differs from the number that identifies the measure. It is fully
independent of `MeasureData::number` ‚Äî `number`'s existing "empty means index + 1" normalization
does not touch it. An empty string is not emitted, since `text=""` is not legal MusicXML and
`core::MeasureText` would repair it to `"-"`; suppressing the printed number is what `implicit`
is for.

`measureNumberingStaffIndex` (`std::optional<int>`) carries `<measure-numbering staff="N">`, the
staff used as the reference point for vertical positioning of the number. Zero-based, converted
across the impl boundary, matching the sibling `measureNumbering*` fields. Absent means the top
staff.

Both fields got `MXAPI_EQUALS_MEMBER` lines.

No `roundtrip-baseline.txt` entries added. Discovery shows `synthetic/measure.3.1.xml` moved from
failing on `text` to failing on the still-unmodeled `<measure id=...>`, and
`synthetic/measure-numbering.4.0.xml` still fails on the unmodeled print-style-align attributes.
Neither pass is available yet, so neither is attributable to this change.

## Testing

- [x] Five new `MeasureData` cases cover round-trip, omission when absent, and empty-string
      suppression: `displayedNumberRoundTrip`, `displayedNumberAbsentOmitsAttribute`,
      `emptyDisplayedNumberOmitsAttribute`, `measureNumberingStaffRoundTrip`,
      `measureNumberingStaffAbsentOmitsAttribute`
- [x] `make api-test` passes (5237 assertions in 468 test cases)
- [x] `make api-roundtrip` passes (295 of 295 pinned files)
- [x] `make fmt-check` passes

### DIFF
--- a/src/include/mx/api/MeasureData.h
+++ b/src/include/mx/api/MeasureData.h
@@ -13,6 +13,7 @@
 #include "mx/api/TransposeData.h"
 
 #include <map>
+#include <optional>
 #include <string>
 
 namespace mx
@@ -51,6 +52,15 @@ class MeasureData
     // about measure numbering can be defined using the measure-numbering element.
     std::string number;
 
+    // The measure number as it should appear on the page, when that differs from the 'number'
+    // above. 'number' identifies the measure (and aligns it with the same measure in other parts),
+    // while this is purely what the engraver prints. Set it when the printed numbering does not
+    // follow the identifying numbering -- a second ending that restarts at 8, an editor's numbering
+    // like "12a", or a rehearsal-driven relabeling. Leave it absent and the printed number is the
+    // 'number' value. An empty string is not written; use implicit = Bool::yes to suppress the
+    // number entirely, which is also what makes engravers ignore this field.
+    std::optional<std::string> displayedNumber;
+
     // The measure-numbering-value type describes how measure numbers are displayed on this part:
     // no numbers, numbers every measure, or numbers every system.
     MeasureNumbering measureNumbering;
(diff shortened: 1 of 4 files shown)


src/include/mx/api/MeasureData.h
```diff
@@ -65,6 +75,10 @@ class MeasureData
     // The <measure-numbering system="..."> attribute; see SystemRelation.
     SystemRelation measureNumberingSystemRelation;
 
+    // Which staff of the part the measure number is vertically positioned against, zero-based from
+    // the top staff. Meaningful only when measureNumbering != unspecified. Absent means the top staff.
```
webern, line 79: Meaningful only when measureNumber != unspecified is not ideal. But there are plenty of patterns already like this and it's easier/faster to just do this than to redesign it now. So I agree with the choice.

webern (approved):
Nice. I agree with a quicker fix now instead of redesigning or refactoring right this moment.

PR #375, opened 2026-07-29
feat: clef additional and after-barline in mx::api

rpatters1:
## Human Summary

Add Bool ClefData::additional and map MusicXML's clef@additional in both directions. unspecified (the default) omits the attribute; yes/no are
written verbatim.

## Summary

Also fix a write-side gap found while adding it: ClefData::location was
already populated by MeasureReader from clef@after-barline, but
PropertiesWriter never emitted the attribute, so the value was silently
lost on write. afterBarline and beforeBarline now write after-barline
yes/no; unspecified and midMeasure emit nothing, since MusicXML ignores the attribute on mid-measure clefs. Document ClefLocation, which had no comment.

Tests cover yes/no/omitted for both attributes, plus an all-fields
round-trip over ClefData that pins the remaining eight fields. That test
records one deliberate normalization: writeStaffNumber=no on a
single-staff part reads back as unspecified, because the knob records
only a divergence from the automatic rule and the automatic rule already
omits the number there. The emitted XML is identical either way.

data/synthetic/clef.3.0.xml and clef.3.1.xml still fail the api
roundtrip on unmodeled position, font, color, and id attributes, so
neither is pinned. The discovery pass-list is unchanged at 297.

### DIFF
--- a/src/include/mx/api/ClefData.h
+++ b/src/include/mx/api/ClefData.h
@@ -21,6 +21,11 @@ enum class ClefSymbol
     jianpu
 };
 
+// Where a clef sits relative to its measure. A clef at the start of a measure normally prints
+// before the barline; beforeBarline and afterBarline choose between the two, which matters for cue
+// clefs and for the clef that resumes after a repeated section. midMeasure describes a clef change
+// partway through the measure, positioned by tickTimePosition. unspecified leaves the choice to the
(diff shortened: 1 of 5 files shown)


webern (approved):
This one is pretty straightforward and looks good. Easy to approve from the phone. 👍

PR #376, opened 2026-07-29
feat: chord degree display options in mx::api

rpatters1:
## Human Summary

Add `parenthesesDegrees` and `stackDegrees` to `ChordData`. Also modified writer to write print-object for Degree. (It was already being read.)

## Summary

Three MusicXML attributes that control how a chord symbol's degrees are displayed were unreachable through `mx::api`.

`ChordData` gains `stackDegrees` and `parenthesesDegrees`, both `api::Bool` defaulting to `unspecified`. They map to the `<kind>` attributes `stack-degrees` (degrees stacked vertically rather than left to right) and `parentheses-degrees` (all degrees wrapped in parentheses, e.g. `C7(#9b13)`). `DirectionReader` reads them off `core::Kind`; `DirectionWriter` emits them only when the field is set, so an author who never touches them gets the same XML as before. The core type already supported both. No behavior change for existing callers.

Separately, `DirectionWriter` was dropping `Extension::printObject`: each `core::Degree` was built with only degree-type, degree-value, and degree-alter, so the `<degree>` `print-object` attribute was never written. The read side already populated the field, so it round-tripped to nothing. MusicXML marks a degree `print-object="no"` when the `<kind>` text attribute already spells it out — Finale exports `<kind text="7sus4">suspended-fourth</kind>` followed by a hidden added seventh, and without the writer support that seventh prints twice.

`data/synthetic/kind.3.0.xml` is now registered with `MxFileRepository` so it can serve as a read fixture. No new corpus files, so no pinned-count or audit regeneration.

## Testing

- [x] `harmonyDegreePrintObjectRoundTrip` fails before the writer fix (2 assertions), passes after
- [x] Six new cases in `HarmonyExtrasApiTest.cpp`: round-trip, negative (unspecified writes no attribute), and read-path coverage for both features
- [x] Read path pinned against real files: `synthetic/kind.3.0.xml` for the kind attributes, and the three hidden degrees in `recsuite/BrookeWestSample.xml` for print-object
- [x] All harmony tests pass (`harmony*`: 32 assertions in 9 test cases)
- [x] Full api suite passes (5242 assertions in 469 test cases)
- [x] `make api-roundtrip` clean (295 passed, 0 failed of 295 pinned)
- [x] `make fmt-check` clean on touched files

Discovery is unchanged. `synthetic/kind.3.0.xml` still fails the strict compare, now on `bracket-degrees` rather than `stack-degrees`; `recsuite/BrookeWestSample.xml` fails earlier in the document on a `credit-words` attribute count. Neither is baseline-eligible yet.

### DIFF
--- a/src/private/mxtest/file/MxFileRepositoy.cpp
+++ b/src/private/mxtest/file/MxFileRepositoy.cpp
@@ -478,5 +478,6 @@ void MxFileRepository::initializeNameSubdirectoryMap()
     myNameSubdirectoryMap.emplace("segno.3.1.xml", "synthetic");
     myNameSubdirectoryMap.emplace("coda.3.0.xml", "synthetic");
     myNameSubdirectoryMap.emplace("coda.3.1.xml", "synthetic");
+    myNameSubdirectoryMap.emplace("kind.3.0.xml", "synthetic");
 }
 } // namespace mxtest
--- a/src/include/mx/api/ChordData.h
+++ b/src/include/mx/api/ChordData.h
@@ -251,6 +251,13 @@ class ChordData
     ChordKind chordKind;
     std::string text;
     Bool useSymbols;
+    // How the chord's degree alterations (the 'extensions' below) are laid out next to the chord
+    // symbol. stackDegrees stacks them vertically instead of running them left to right;
+    // parenthesesDegrees wraps all of them in parentheses, e.g. C7(#9b13). These are the MusicXML
+    // stack-degrees and parentheses-degrees attributes; unspecified leaves the choice to the
+    // notation software rendering the score.
+    Bool stackDegrees;
+    Bool parenthesesDegrees;
     Step bass;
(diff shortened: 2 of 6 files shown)


src/private/mxtest/file/MxFileRepositoy.cpp
```diff
@@ -478,5 +478,6 @@ void MxFileRepository::initializeNameSubdirectoryMap()
     myNameSubdirectoryMap.emplace("segno.3.1.xml", "synthetic");
     myNameSubdirectoryMap.emplace("coda.3.0.xml", "synthetic");
     myNameSubdirectoryMap.emplace("coda.3.1.xml", "synthetic");
+    myNameSubdirectoryMap.emplace("kind.3.0.xml", "synthetic");
```
webern, line 481: Nice.

webern (approved):

PR #377, opened 2026-07-29
feat: add notehead filled and smufl to mx::api

rpatters1:
## Human Summary

The title says it all. Add notehead filled and smufl to mx::api. Also cleanup as identified by Claude.

## Summary

The `<notehead>` element's `filled` and `smufl` attributes were modeled in `mx::core` but dropped by `mx::api` in both directions. `NoteData` now carries them alongside `notehead`:

- `noteheadFilled` is an `api::Bool` defaulting to unspecified, so the attribute is only written when set and MusicXML's duration-based default (hollow for a half note and longer) otherwise applies.
- `noteheadSmufl` is an optional glyph name, following the precedent in `PercussionData` and `OtherDirectionData`. An empty string is treated as absent rather than letting `core::SmuflGlyphName` repair it into the placeholder "X".

`NoteWriter::setNotehead` previously skipped the element entirely whenever the notehead value was
`normal`. It now emits when either attribute is set, so `<notehead filled="no">normal</notehead>`
survives a round trip. `NoteReader::setNoteheadValue` becomes `setNoteheadItems` and reads all
three off `core::Notehead`; `NoteFunctions::parseNote` copies them onto the api struct.

Also adds the missing `MXAPI_EQUALS_MEMBER(notehead)`. That field had never been in `NoteData`'s
equality block, so two notes with different noteheads compared equal and the field was invisible
to round-trip checks.

No `roundtrip-baseline.txt` additions. `lysuite/ly22a_Noteheads.xml` gets past its notehead
attribute mismatch but still fails further along on three explicit `<notehead>normal</notehead>`
elements the writer drops, which needs a separate fidelity knob. The synthetic notehead files now
fail on `parentheses`, font and color, which mx::api does not model.

## Testing

- [x] Six new cases in NoteDataTest: filled alone, filled with a shape, smufl with
      `Notehead::other`, both together on `cluster`, absence by default, and a read of
      `data/synthetic/notehead.3.1.xml` that pins the core to api path
- [x] All notehead tests pass (`notehead*`: 38 assertions in 9 test cases)
- [x] All NoteData tests pass (`*_NoteData`: 285 assertions in 37 test cases)
- [x] Full api suite passes (5249 assertions in 469 test cases)
- [x] `make api-roundtrip` passes (295 of 295 pinned)

### DIFF
--- a/src/include/mx/api/NoteData.h
+++ b/src/include/mx/api/NoteData.h
@@ -13,6 +13,7 @@
 #include "mx/api/PrintData.h"
 
 #include <optional>
+#include <string>
 #include <vector>
 
 namespace mx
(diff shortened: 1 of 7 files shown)


webern (approved):
Nice, thank you for all these contributions.

You can merge them in the order you prefer. Only one of them is not approved yet, the one that I mentioned seems to not have the desired diff at the moment.

rpatters1:
Just a reminder: I don't have merge privilege (and am not asking for it).

rpatters1:
These should be mergeable in any order. (That was the intent.) But the safest order would be from least-to-most recent.

PR #378, opened 2026-07-30
feat: extend direction text formatting in mx::api and unify the enclosure enum

rpatters1:
## Human Summary

Extends direction text formatting with `justify` and `systemRelation` added where available in the musicxml spec. Also unified the `Enclosure` enum which resulted in a breaking change, but it seems like the best approach moving forward. (See below for more details.)

## Summary

Four MusicXML attributes that mx::api could not express are now modeled, and the two duplicate enclosure enums are merged into one.

justify on words, symbol and rehearsal. All three elements carry both justify and halign, which MusicXML treats as different things: halign says which edge of the text the position refers to, justify says how the lines sit relative to each other when the text runs to more than one line. halign was already reaching the api through positionData.horizontalAlignment; justify was dropped. PageTextData and TempoData already had a justify field, so the new ones follow that precedent. Elements carrying justify without halign (lyric, part-name, group-name) are deliberately left alone -- the spec says justify doubles as alignment there, which is how the lyric reader already treats it.

systemRelation on DirectionData, mapping the system attribute of both direction and harmony. The api::SystemRelation enum already existed for measure numbering. The direction vocabulary is narrower (only-top, also-top, none), so the api-to-core conversion returns an optional and writes no attribute for the two measure-numbering-only values instead of coercing them into something wrong.

The enclosure vocabulary is now complete. RehearsalEnclosure was missing invertedBracket and pentagon through decagon, while PercussionEnclosure already had all fifteen. Since both described the same MusicXML enclosure-shape type with identical values, they are replaced by a single Enclosure in ApiCommon.h. That also collapses three parallel conversion paths -- two hand-written switches in DirectionReader and DirectionWriter plus the Converter table -- into the Converter table alone, which is where the drift came from in the first place.

**Breaking change**: `RehearsalEnclosure` and `PercussionEnclosure` are removed rather than aliased. The vocabularies are unchanged, so the downstream fix is a mechanical rename to `Enclosure`. Otherwise there is no structural API change; the additions are new fields on existing types with defaults that preserve current behavior.

## Testing

- [x] 11 new tests across DirectionMarksRoundTripTest, DirectionDataTest and HarmonyExtrasApiTest, covering each attribute set and left unspecified, the unwritable system values, and every shape in the enclosure vocabulary
- [x] make api-test: 5367 assertions in 474 test cases
- [x] make api-roundtrip: 295 passed, 0 failed of 295 pinned
- [x] make core-roundtrip-test: 837 test cases
- [x] Unity build (CMAKE_UNITY_BUILD_BATCH_SIZE=0), since two file-local functions were removed

No corpus fixture was pinned to roundtrip-baseline.txt: the files exercising these attributes (synthetic/direction.4.0.xml, synthetic/words.3.x.xml) still fail the api round trip on unrelated gaps, directive on direction and dir on words.

### DIFF
--- a/src/include/mx/api/ApiCommon.h
+++ b/src/include/mx/api/ApiCommon.h
@@ -71,9 +71,11 @@ enum class MeasureNumbering
     system
 };
 
-// The <measure-numbering system="..."> attribute: whether this part's measure numbers are also,
-// or only, associated with the system (as opposed to just this part). unspecified means the
-// attribute is absent.
+// Whether an item is associated with the system rather than only with the part it appears in --
+// the system attribute of <measure-numbering> (MeasureData) and of <direction> and <harmony>
+// (DirectionData). only... means the item is drawn on the top or bottom part of the system instead
+// of this part; also... means it is drawn on both. unspecified means the attribute is absent.
+// onlyBottom and alsoBottom are measure-numbering only; MusicXML has no bottom-of-system direction.
 enum class SystemRelation
 {
     unspecified,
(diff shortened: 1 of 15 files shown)


rpatters1:
This is the same as #374 but based on the correct source branch.

src/include/mx/api/ApiCommon.h
```diff
@@ -71,9 +71,11 @@ enum class MeasureNumbering
...
 };
 
-// The <measure-numbering system="..."> attribute: whether this part's measure numbers are also,
-// or only, associated with the system (as opposed to just this part). unspecified means the
-// attribute is absent.
+// Whether an item is associated with the system rather than only with the part it appears in --
+// the system attribute of <measure-numbering> (MeasureData) and of <direction> and <harmony>
+// (DirectionData). only... means the item is drawn on the top or bottom part of the system instead
+// of this part; also... means it is drawn on both. unspecified means the attribute is absent.
+// onlyBottom and alsoBottom are measure-numbering only; MusicXML has no bottom-of-system direction.
```
webern, line 78: I think this is unfortunate, but I don't disagree with the choice you made. There's often a tradeoff like this when trying to simplify the MusicXML surface area.

webern (approved):
Looks good. I think having enums where certain members are only applicable in certain scenarios is not ideal, but you are not wrong to choose this simplification. I think I balked at it at first which is why it took longer to review. Sorry for the delay.

PR #382, opened 2026-08-02
feat: support SMuFL other marks in mx::api

### LINKED ISSUE #381: Add SMuFL payloads for MusicXML other-* marks and expose other-notation
## Problem

MusicXML provides several extension elements for notation that does not have a dedicated element:

- `<other-dynamics>`
- `<other-articulation>`
- `<other-technical>`
- `<other-ornament>`
- `<other-notation>`

Each of these elements can carry a `smufl` attribute naming one specific SMuFL glyph. This lets applications exchange the precise symbol while optionally retaining text content as a fallback or description.

`mx::core` models these elements and their SMuFL attributes, but `mx::api` does not expose that information. For the first four elements, `MarkData::name` carries the element text while the reader drops `smufl` and the writer cannot emit it. `<other-notation>` is not exposed at all: `NoteFunctions` ignores it during import and the API writer has no corresponding representation.

This prevents a consumer from preserving or authoring a precise fallback component such as the `z` in an `ffz` dynamic:

```xml
  <other-dynamics smufl="dynamicZ">z</other-dynamics>
```

The same loss occurs when a note-attached symbol is only representable as an `other-articulation`, `other-technical`, `other-ornament`, or general `other-notation`.

## Ordered dynamic components

Although one `smufl` attribute names only one glyph, MusicXML's `<dynamics>` type contains an unbounded ordered choice of dynamic children. This is how a mark such as `ffz`, for which MusicXML has no dedicated `<ffz/>` element, can combine a classified dynamic with a SMuFL fallback:

```xml
<dynamics>
  <ff/>
  <other-dynamics smufl="dynamicZ">z</other-dynamics>
</dynamics>
```

The same mark may instead be represented entirely as individually named glyph components:

```xml
<dynamics>
  <other-dynamics smufl="dynamicFF">ff</other-dynamics>
  <other-dynamics smufl="dynamicZ">z</other-dynamics>
</dynamics>
```

`mx::core::Dynamics` preserves this ordered sequence. The current API translation does not: `DynamicsReader` examines only `choice().front()`, and `DynamicsWriter` constructs one dynamics choice from one `MarkData`. Consequently, `mx::api` cannot preserve a multi-component `<dynamics>` mark even apart from the missing `smufl` attribute.

## Context

This work was intentionally deferred during the mx::impl port; see [the SMuFL glyph-name migration notes](docs/ai/design/mx-impl-port-plan.md#12-deferred-to-phase-3-the-smufl-glyph-name-migration-open-questions). That section records the pre-MusicXML-4.0 workaround of putting canonical glyph names in element text, notes that the generated core has the proper `smufl` attribute, and leaves the public-API and compatibility policy for later design.

The deferral appears to have been about sequencing the port rather than a decision that the feature should remain unsupported. There is now a concrete downstream need: exporters must currently downgrade precise Finale articulation and dynamic glyphs to text-valued `other-*` elements, split or flatten compound dynamics, or substitute less accurate semantic notation.

The existing design questions remain relevant:

- How should SMuFL glyph names be represented in `mx::api` without adding fields that are meaningless for most marks?
- How should the API represent the ordered children of one `<dynamics>` element, including mixtures of semantic and `other-dynamics` components?
- Should legacy element text that looks like a SMuFL name ever be promoted automatically, or should migration remain the consumer's responsibility?
- How should fallback text and a canonical SMuFL name coexist for compatibility with readers that ignore the attribute?
- Should the legacy `customAccentTenuto` sentinel path be retained, migrated, or handled separately?

## Desired outcome

Design and implement an `mx::api` representation that can read and write:

- the SMuFL glyph name on the existing `other-*` mark categories;
- the ordered components of a MusicXML `<dynamics>` mark; and
- MusicXML's general `<other-notation>` element.

The result should:

- preserve optional SMuFL glyph names through XML -> API -> XML;
- preserve the number, kind, and order of children within one `<dynamics>` element;
- support API-authored glyph-only elements as well as elements containing fallback text;
- preserve the meaningful `other-notation` attributes supported by `mx::core`;
- remain compatible with existing API-authored `other-*` marks that do not supply a SMuFL glyph;
- define an explicit policy for legacy glyph names stored in element text; and
- include unit and corpus round-trip coverage using the existing `data/synthetic/other-*.3.0.xml` and `other-*.3.1.xml` fixtures where applicable, plus a fixture containing multiple dynamic children.

rpatters1:
## Human Summary

Preserve SMuFL glyph names for MusicXML `other-*` marks and expose general `other-notation` through `mx::api`. Compound dynamics such as `ffz` now retain all ordered components instead of losing everything after the first child.

## Summary

Add mark-specific API payloads for `other-articulation`, `other-dynamics`, `other-ornament`, and `other-technical`, keeping fallback text and the optional `smufl` glyph name independent.

Represent one MusicXML `<dynamics>` element as an ordered compound mark when it contains multiple children. Single standard dynamics retain the existing `MarkData` representation for source compatibility.

Add symmetric `other-notation` support for text, type, number, `smufl`, id, position, and print data. The change also fixes explicit `print-object="no"` emission and retires anonymous namespaces in touched files for unity-build compatibility.

Nine existing synthetic `other-*` fixtures now pass and are pinned in the API round-trip baseline.

## Testing

- [x] `make api-test MX_RUNNING_IN_DOCKER=1` (500 test cases, 5510 assertions)
- [x] `make api-roundtrip MX_RUNNING_IN_DOCKER=1` (305 pinned files, 0 failed)
- [x] `make fmt-check MX_RUNNING_IN_DOCKER=1`
- [x] Strict unity build of target `mx`
- [x] `git diff --check`

## References

- Closes #381


src/include/mx/api/MarkData.h
```diff
@@ -73,6 +73,7 @@ enum class MarkType
     pf,
     sfzp,
```
webern: Should all of this be replaced by standardDynamics so that we don't have duplication of the standard dynamic marks? Not sure, but the duplication of these into DynamicsData.h feels strange.

rpatters1: Okay, I'm waffling a bit. We can get rid of StandardDynamic, but then we have to use MarkType for other dynamics. This is fine, but it allows nonsense other dynamics arrays like `{accent, mf}`. I have a commit ready to go that gets rid of `StandardDynamic` and does its best to enforce that no dynamic gets so-initialized. But it isnt as ironclad as a dedicated enum. Your call.

rpatters1:
Okay, I refactored so that dynamics types are removed from `MarkDataType` and now live in `StandardDynamic`. This is therefore a breaking change, but it is a loud breaking change that can be found by compiling. I iterated with Claude a good bit on this one, and I think where we ended up is a very solid representation. A lot of code got collapsed.

webern (approved):
Nice. I think it's better with the deduplication even if it's a bigger breaking change. 👍

PR #384, opened 2026-08-03
feat: artificial and natural harmonic detail in mx::api

rpatters1:
## Human Summary

mx::api could only ever write an undecorated `<harmonic/>`. The natural/artificial distinction and the base/touching/sounding-pitch detail were dropped in both directions, so an artificial harmonic was indistinguishable from a natural one. This adds them as a MarkDataChoice payload.

Also fixes the harmonic writer dropping print-object, font, and color. The wider print-object gap it uncovered is called out in the Summary and left for a separate change.

## Summary

MusicXML's `<technical><harmonic>` states how a harmonic is produced (`<natural/>` or `<artificial/>`) and which of its three pitches the written notehead represents (`<base-pitch/>`, `<touching-pitch/>`, `<sounding-pitch/>`). `mx::core::Harmonic` models both, as `choice` and `choice2`, but `mx::api` dropped them in both directions: `NotationsWriter::addTechnical` built a bare `core::Harmonic` and never called `setChoice`/`setChoice2`, and `TechnicalFunctions::parseTechicalMark` read only the attributes.

Following the `TremoloMarkData` / `ArpeggiateMarkData` precedent, this adds a fourth `MarkDataChoice` alternative:

- `HarmonicMarkData::kind` is a `HarmonicKind` of unspecified, natural, or artificial.
- `HarmonicMarkData::pitch` is a `HarmonicPitch` of unspecified, basePitch, touchingPitch, or soundingPitch.

Both default to unspecified, so authoring without the payload writes the plain `<harmonic/>` exactly as before, and the reader leaves `choice` as none for a bare element rather than inventing a natural/artificial value. The two notes of an artificial harmonic each carry their own mark, artificial with basePitch on the stopped note and artificial with touchingPitch on the diamond-notehead note, so neither note depends on its neighbor for meaning.

Also fixes a write-side gap found while adding it. The harmonic writer used `setAttributesFromPositionData`, so the element's font, color, and print-object were all lost on write. It now uses `setAttributesFromMarkData` and sets print-object explicitly, because `setAttributesFromPrintData` writes color and font but never touches printObject. That omission still affects every other mark going through `setAttributesFromMarkData` (fermata, dynamics, arpeggiate, non-arpeggiate), and the `impl::setPrintObject` template in `PrintFunctions.h` that looks like it was meant to cover it is unused and has its condition inverted. Left for a separate change.

`data/synthetic/harmonic.3.0.xml` already existed but had never passed; it is now pinned (296 -> 297). `lysuite/ly32a_Notations.xml` and `ksuite/k004a_Technical.xml` get past their harmonic content but still fail on unrelated `<lyric>` and `<alter>` ordering, so neither is pinned.

## Testing

- [x] Three new cases in MarkRoundTripTest: the artificial pair (base pitch and touching pitch), all eleven kind/pitch combinations, and a bare `<harmonic/>` that must read back with no payload
- [x] Harmonic tests pass (`*Harmonic*`: 43 assertions in 3 test cases)
- [x] Full api suite passes (5512 assertions in 499 test cases)
- [x] `make api-roundtrip` passes (297 of 297 pinned)
- [x] `make core-roundtrip-test` passes (837 test cases)

## References

- Related to #382 and #383, which also add `MarkDataChoice` alternatives. Whichever of the three lands first, the others need a union resolution in `MarkDataChoice.h` and `MarkDataChoice.cpp`; the three payloads are independent.

### DIFF
--- a/src/include/mx/api/MarkDataChoice.h
+++ b/src/include/mx/api/MarkDataChoice.h
@@ -137,6 +137,63 @@ MXAPI_EQUALS_MEMBER(id)
 MXAPI_EQUALS_END;
 MXAPI_NOT_EQUALS_AND_VECTORS(OtherNotationMarkData);
 
+// How a harmonic is produced on a string instrument.
+enum class HarmonicKind
+{
+    // Leave the production method unstated, drawing only the circular harmonic symbol.
+    unspecified,
+
+    // The string sounds open and is touched lightly at a node, so the harmonic's pitch is fixed by
+    // where the node falls on the open string.
+    natural,
+
+    // The string is stopped at one pitch and touched lightly a fixed interval above it -- usually a
+    // fourth, sometimes a major third or a fifth. The stopped note is written as an ordinary
+    // notehead and the touched note as a diamond (see NoteData::notehead).
+    artificial,
+};
+
+// Which of a harmonic's three pitches the note's written notehead states.
+enum class HarmonicPitch
+{
+    // Leave it unstated. Notation that writes only one note for the harmonic, or that relies on
+    // notehead shape alone to say which pitch is meant, does not need this.
+    unspecified,
+
+    // The stopped pitch: where the finger presses the string. The lower note of an artificial
+    // harmonic pair.
+    basePitch,
+
+    // The lightly touched pitch. The upper, diamond-notehead note of an artificial harmonic pair,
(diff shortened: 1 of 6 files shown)


webern:
I'm getting caught up at work post vacation and will get on these PRs soon!

webern (approved):

PR #392, opened 2026-08-07
feat: add emscripten build to ci

### LINKED ISSUE #386: Add emscripted build to CI
A powerful feature of this library (and the musx -> musicxml library I am building on top of it) is the ability to compile it with `emscripten` and use it client-side in a browser from a website.

If the CI is going to be refactored for Windows, I think including an emscripten build is something to consider as well. Right now denigma is building for emscripten in a separate repo, but I may add that test to the main repo at some point.

_Originally posted by @rpatters1 in https://github.com/webern/mx/discussions/355_

webern:
## Human Summary

Add a smoke build to WASM and run something trivial to make sure we don't break [denigma](https://github.com/rpatters1/denigma).

## Summary

`mx` already gets cross-compiled to WebAssembly downstream today: `denigma` (github.com/rpatters1/denigma) builds it client-side for browser use in a musx-to-MusicXML converter, but nothing here proves that keeps working, so a regression would only surface in someone else's CI. This adds an Emscripten build to `mx`'s own CI.

- `mx-sdk`'s `Dockerfile` gets a pinned `emsdk` install (6.0.6), matching how every other tool in that image is version-pinned rather than "latest". `EM_CACHE` is redirected to the mounted build volume, same reasoning as `CCACHE_DIR`/`GOPATH` right below it in the file: the container runs as the caller's uid:gid, which can't write under the root-owned `/opt/emsdk` at runtime.
- `CMakeLists.txt` now enables exception catching under `EMSCRIPTEN` (`-fexceptions`, compile and link). Emscripten disables catching by default, which turns `mx`'s internal `MX_THROW`/catch boundary in `DocumentManager` (`src/private/mx/utility/Throw.h`) into a fatal abort instead of a caught `ResultCode::internalError`. Previously only `denigma`'s own build set this flag (globally, before `FetchContent`-ing `mx`) -- `mx` didn't configure itself correctly for `EMSCRIPTEN` on its own. Confirmed this is the same fix `denigma-examples`' top-level `CMakeLists.txt` already applies, so this reproduces a known-working setup rather than a new one.
- New `make wasm-lib` / `wasm-build` / `wasm-test` targets, driven inside `mx-sdk` like every other gate. `wasm-test` builds `mx::api` and runs the existing `mxread`/`mxwrite`/`mxhide` examples under Node -- it exercises the library through `DocumentManager`, not just a compile.
- New `test-wasm` CI job (mirrors `test-linux`/`quality`/`gen`'s Docker setup) running `make wasm-test`.

## Testing

- [x] `make -n wasm-test` (dry run, both outside- and inside-container branches) resolves to the expected `emcmake`/`emmake`/`node` commands
- [x] Native `cmake -S . -B ... -DMX_API=on` still configures cleanly with the new `EMSCRIPTEN` block present (a no-op when not cross-compiling)
- [x] `ci.yaml` still parses as valid YAML
- [x] CI

## References

- Closes #386

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -96,6 +96,42 @@ jobs:
       - name: All C++ suites (core roundtrip + unit, api mxtest + examples + roundtrip)
         run: make test-all
 
+  # mx::api cross-compiled to WebAssembly via the pinned mx-sdk emsdk install
+  # (issue #386): a consumer (denigma, github.com/rpatters1/denigma) compiles mx
+  # client-side for browser use, and this is the one place that path is proven
+  # rather than discovered as a regression downstream. Builds through the
+  # mx-sdk toolchain like test-linux/quality/gen; runs mxread/mxwrite/mxhide
+  # under Node so the check exercises mx::api, not just a compile.
+  test-wasm:
+    name: "wasm: mx::api under Emscripten"
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v5
+
+      - uses: docker/setup-buildx-action@v3
+      - uses: crazy-max/ghaction-github-runtime@v3
+
+      - name: Create bind-backed build volume
+        run: |
+          mkdir -p build/docker
+          docker volume create --driver local \
+            --opt type=none --opt o=bind \
+            --opt device="$GITHUB_WORKSPACE/build/docker" mx-build
+
+      # Emscripten's compiled system-library cache (libc/libc++/compiler-rt),
+      # not source ccache -- see EM_CACHE in the Dockerfile. Populating it from
+      # scratch is the slow part of a wasm build.
+      - name: Cache EM_CACHE
+        uses: actions/cache@v4
+        with:
+          path: build/docker/.emcache
+          key: emcache-${{ runner.os }}-wasm-${{ github.sha }}
+          restore-keys: emcache-${{ runner.os }}-wasm-
+
+      - name: mx::api under Node (wasm)
+        run: make wasm-test
+
   # The api product surface, built natively with AppleClang. macOS is a
   # portability check on the product tier only (the core suites run on Linux in
   # test-linux); this is the only C++ job off the pinned toolchain, by design,
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -97,11 +97,7 @@ jobs:
         run: make test-all
 
   # mx::api cross-compiled to WebAssembly via the pinned mx-sdk emsdk install
-  # (issue #386): a consumer (denigma, github.com/rpatters1/denigma) compiles mx
-  # client-side for browser use, and this is the one place that path is proven
-  # rather than discovered as a regression downstream. Builds through the
-  # mx-sdk toolchain like test-linux/quality/gen; runs mxread/mxwrite/mxhide
-  # under Node so the check exercises mx::api, not just a compile.
+  # (#386) -- denigma (rpatters1/denigma) depends on this path staying green.
   emscripten:
     name: emscripten
     runs-on: ubuntu-latest
@@ -119,10 +115,8 @@ jobs:
             --opt type=none --opt o=bind \
             --opt device="$GITHUB_WORKSPACE/build/docker" mx-build
 
-      # Two caches: ccache for mx's own object files (emcc via the ambient
-      # CMAKE_*_COMPILER_LAUNCHER), EM_CACHE for Emscripten's compiled
-      # system libraries (libc/libc++/compiler-rt). Without both, this job
-      # recompiles everything from scratch on every run.
+      # ccache for mx's own object files; EM_CACHE (below) is Emscripten's
+      # separate cache for its compiled system libraries.
       - name: Cache ccache
         uses: actions/cache@v4
         with:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -102,8 +102,8 @@ jobs:
   # rather than discovered as a regression downstream. Builds through the
   # mx-sdk toolchain like test-linux/quality/gen; runs mxread/mxwrite/mxhide
   # under Node so the check exercises mx::api, not just a compile.
-  test-wasm:
-    name: "wasm: mx::api under Emscripten"
+  emscripten:
+    name: emscripten
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -35,6 +38,36 @@ RUN python3 -m venv /opt/quality-venv \
...
+# emsdk: the Emscripten toolchain, used to build mx to WebAssembly (issue #386;
+# see CMakeLists.txt's EMSCRIPTEN block). Pinned to a specific release, not
+# "latest", for the same reproducibility reason every other tool in this image
+# is version-pinned. emsdk writes its config (.emscripten) inside $EMSDK by
+# default (has since 1.39.x) rather than $HOME, so it needs no help from us
+# there -- but its system-library cache defaults to living under $EMSDK too,
+# which the container's caller-uid:gid user (no passwd entry, root-owned
+# /opt/emsdk) can't write to at runtime; EM_CACHE below redirects that.
+# node's bundled bin dir has a version number baked into its path; symlink a
+# stable name so PATH below doesn't have to guess it.
```
webern: Wayyyyy to much, dude. Scale it back to one or two lines.

Dockerfile
```diff
@@ -35,6 +38,36 @@ RUN python3 -m venv /opt/quality-venv \
...
+    && rm -rf /opt/emsdk/downloads
+
+# EM_CACHE (the compiled libc/libc++/compiler-rt system-library cache) is
+# redirected to the build volume for the same reason CCACHE_DIR is below: it
+# would otherwise try to write under the root-owned /opt/emsdk at runtime.
+# PATH appends emsdk's dirs rather than prepending them: upstream/emscripten
+# (Emscripten's own tree) contains a subdirectory literally named `cmake`
+# (its CMake toolchain-file module), and execve() on a directory returns
+# EACCES ("Permission denied") -- prepended, it silently shadowed the real
+# /usr/bin/cmake for every job, not just the wasm one.
```
webern: Wayyyyy to much, dude. Scale it back to one or two lines.

Makefile
```diff
@@ -235,6 +241,35 @@ api-coverage:
 		$(BUILD_ROOT)/cov-api | tee $(COV_DIR)/api/summary.txt
 	@echo "=== api-coverage written to $(COV_DIR)/api/ ==="
 
+# wasm: mx::api built for Emscripten (see CMakeLists.txt's EMSCRIPTEN block for
+# the exceptions flag this relies on). Compiler launcher blanked, same as the
+# coverage targets above -- ccache wrapping emcc's Python driver is untested
+# and not worth the risk for a job that isn't anyone's hot inner loop.
```
webern: Wayyyyy to much, dude. Scale it back to one or two lines.

Makefile
```diff
@@ -235,6 +241,35 @@ api-coverage:
...
+	emmake $(CMAKE) --build $(BUILD_ROOT)/wasm --target mxread mxwrite mxhide --parallel $(JOBS)
+
+# Proves mx::api actually works under wasm, not just that it compiles:
+# mxread/mxwrite/mxhide (src/private/mx/examples/) round-trip real ScoreData
+# through DocumentManager, the same programs api-examples runs natively.
+# Emscripten's CMake toolchain names the outputs *.js; node runs them and
+# each one's C++ main() return code becomes node's exit code. mxwrite's
+# writeToFile lands in Emscripten's default in-memory MEMFS, not on the real
+# filesystem -- the output path won't actually appear under build/wasm/, only
+# the write call (and its return code) is being exercised here.
```
webern: Wayyyyy to much, dude. Scale it back to one or two lines.

CMakeLists.txt
```diff
@@ -12,6 +12,17 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 set(EXECUTABLE_OUTPUT_PATH ${CMAKE_BINARY_DIR})
 set(LIBRARY_OUTPUT_PATH ${CMAKE_BINARY_DIR})
 
+# Emscripten disables exception catching by default, which turns mx's internal
+# MX_THROW (see src/private/mx/utility/Throw.h) into a fatal abort instead of
+# something DocumentManager can catch and turn into ResultCode::internalError.
+# Set directory-scoped (not per-target) so it also covers pugixml/mx_core/mx,
+# and set it here rather than relying on downstream consumers (e.g. denigma)
+# to remember it before FetchContent-ing mx.
```
webern: Wayyyyy to much, dude. Scale it back to one or two lines.

.github/workflows/ci.yaml
```diff
@@ -96,6 +96,50 @@ jobs:
       - name: All C++ suites (core roundtrip + unit, api mxtest + examples + roundtrip)
         run: make test-all
 
+  # mx::api cross-compiled to WebAssembly via the pinned mx-sdk emsdk install
+  # (issue #386): a consumer (denigma, github.com/rpatters1/denigma) compiles mx
+  # client-side for browser use, and this is the one place that path is proven
+  # rather than discovered as a regression downstream. Builds through the
+  # mx-sdk toolchain like test-linux/quality/gen; runs mxread/mxwrite/mxhide
+  # under Node so the check exercises mx::api, not just a compile.
```
webern: two lines max

.github/workflows/ci.yaml
```diff
@@ -96,6 +96,50 @@ jobs:
...
+        run: |
+          mkdir -p build/docker
+          docker volume create --driver local \
+            --opt type=none --opt o=bind \
+            --opt device="$GITHUB_WORKSPACE/build/docker" mx-build
+
+      # Two caches: ccache for mx's own object files (emcc via the ambient
+      # CMAKE_*_COMPILER_LAUNCHER), EM_CACHE for Emscripten's compiled
+      # system libraries (libc/libc++/compiler-rt). Without both, this job
+      # recompiles everything from scratch on every run.
```
webern: three lines max

Makefile
```diff
@@ -235,6 +241,27 @@ api-coverage:
...
+		-DMX_API=on
+	emmake $(CMAKE) --build $(BUILD_ROOT)/wasm --target mx --parallel $(JOBS)
+
+wasm-build: wasm-lib
+	emmake $(CMAKE) --build $(BUILD_ROOT)/wasm --target mxread mxwrite mxhide --parallel $(JOBS)
+
+# Runs mxread/mxwrite/mxhide under Node so this exercises mx::api, not just a
+# compile. mxwrite's output path is bare (no directory) and lands in
+# Emscripten's in-memory MEMFS, not disk -- MEMFS starts empty, so a path
+# with a build/wasm/ prefix would fail outright with no such directory.
```
webern: two lines max

Issue #399, opened 2026-08-15
Make ID attributes available in `mx::api`.


webern:
Beginning with MusicXML 3.1, many MusicXML nodes can have an ID attribute. `mx::api` should make this available. This is related to, and requested in discussion #349 which lead to (the challenging to design and implement) issue #393.

In the discussion @rpatters1 said:

> As an aside, if ID values were available in mx::api, I would be using them a lot more right now.

This is very much in this library's wheelhouse. A bit more complicated is the related #397. But with-or-without uniqueness validation, adding attribute fields that `mx::api` is agnostic about is easy.

Edit:

#399 exposes the most easily reachable IDs, but the following require more research or design:
- [ ] #403 
- [ ] #404 
- [ ] #405 
- [ ] #406 
- [ ] #407 
- [ ] #408

Issue #397, opened 2026-08-15
ID uniqueness is not constrained by `mx::core`.


webern:
`mx::core` does not validate that `xs:ID` values are unique at the document level as required by the XSD specification. Every id is modeled as `Token` (`src/private/mx/core/generated/Note.h:113`, `ScorePart.h:38`). There is no document-wide uniqueness table.

Solving this requires a decision about whether parsing a document that violates this should fail with an error or have some graceful recovery (which probably requires a reporting mechanism). Furthermore, when building a document, how would we handle a violation when an attribute is being added to the document that violates the uniqueness table. Heretofore, our aversion to non-infallible functions has not been too problematic, but with this and #393, we are straying into areas where the document can be more easily screwed up. An immediate error response for trying to insert a duplicate ID into the score is probably a better interface than an error later on a write or validate function.

Edit: an LLM session has swayed me toward an error-on-write instead of fail-fast.

Issue #393, opened 2026-08-14
Support spanners specified as a single object instead of start and stop pairs.


webern:
The idea is that when authoring a document with `mx::api`, it should not be necessary to maintain a spanner state machine so that one can place start and stop sentinels in the score for things like wedges, glissandos, etc. Instead one should be able to specify that spanner as a single unit that says where it starts and stops.

Status: in design on the `spanners` branch. The design will be posted here when finalized.

_Originally posted by @rpatters1 in https://github.com/webern/mx/discussions/349_

**This is a tracking issue:**

- [ ] #397 
- [ ] TBD, add more...

Not sub-issues, but related:
- #372 
- #291 
- #389
- #411
- #412

Issue #389, opened 2026-08-06
api: model note-attached paired spanners (glissando, slide, wavy-line)


rpatters1:
## What

MusicXML attaches three paired spanners to notes inside `<notations>`:

- `<glissando type="start|stop">` — the discrete case; defaults to a wavy line
- `<slide type="start|stop">` — continuous between the two pitches; defaults to a solid line
- `<ornaments><wavy-line type="start|continue|stop">` — trill extensions and vibrato lines

`mx::api` models none of them as spanners.

**Glissando and slide are dropped on read and cannot be authored.** `mx::core` models both
elements and `NotationsChoice` admits them, but `NoteFunctions::parseNotations` has an empty
`break` for `Kind::glissando` and a commented-out body for `Kind::slide`
(`src/private/mx/impl/NoteFunctions.cpp:249-257`), with a dead `SlideFunctions.cpp` stub behind it.
On the write side `CurveType` offers only `slur` and `tie`
(`src/include/mx/api/CurveData.h:45-50`), and `NotationsWriter` skips any entry in `curveStarts`,
`curveContinuations`, or `curveStops` whose type is neither (`NotationsWriter.cpp:112, 133, 154`),
so the generic curve path cannot carry them either. `NoteAttachmentData` has commented-out
`spannerStarts` / `spannerStops` vectors (`src/include/mx/api/NoteAttachmentData.h:24-25`), but
they are not a design in waiting: they are typed `std::vector<SpannerData>`, a flat
enum-discriminated struct that was deliberately replaced by today's `SpannerStart` / `SpannerStop`
in 781a461 (2016). The vectors referenced the deleted type and were commented out rather than
retyped. `SlideFunctions::parseSlide` is the same story — added already commented out, still naming
the long-dead `api::SpannerType::slide`.

**Wavy lines lose their pairing.** `MarkType::wavyLine` reads and writes as an ordinary ornament
mark carrying position and print data only. `core::WavyLine::type()` is a *required*
`StartStopContinue` that the writer never sets (`NotationsWriter.cpp:698-703`), so it falls back to
its default tag, `start`. Every wavy line mx writes therefore comes out `type="start"`: a trill
extension read in as a start/stop pair is written back as two starts, and the line never closes.
`number` and `smufl` are neither read nor written. The reader is symmetric —
`OrnamentsFunctions.cpp:128-131` keeps position and print and discards the rest.

## Why

Glissandi, slides, and trill/vibrato extensions are ordinary notation, not edge cases. Today an
[...]

PR #401, opened 2026-08-18
feat: expose musicxml id attributes in mx::api

webern:
## Human Summary

Model ID for those cases where the mx::api data model is close enough the MusicXML data model to make that simple.

Caveats:
- ID strings are silently coerced into valid NCNames.
- ID uniqueness is not yet enforced.

## Summary

Starting with MusicXML 3.1 most elements can carry an `id` attribute naming that one element within the document. `mx::api` dropped nearly all of them (only a handful of recently-added direction types had one). This adds an optional `id` member to every api type that maps one-to-one onto an element with an id, and reads and writes it in `mx::impl`.

Types that gained an `id`:

| Where | MusicXML element |
|---|---|
| `NoteData`, `LyricData` | note, lyric |
| `MeasureData`, `BarlineData` | measure, barline |
| `ClefData`, `KeyData`, `TimeChoice`, `TransposeData` | clef, key, time, transpose |
| `DirectionData`, `SoundData` | direction, sound |
| `WordsData`, `SymbolData`, `RehearsalData` | words, symbol, rehearsal |
| `WedgeStart`, `WedgeStop` | wedge |
| `SpannerStart`, `SpannerStop` | bracket, dashes, octave-shift |
| `PedalLineData` | pedal |
| `CurveStart`, `CurveContinue`, `CurveStop`, `TieLetRing` | slur, tied |
| `TupletStart`, `TupletStop` | tuplet |
| `FiguredBassData` | figured-bass |

Supporting changes:

- New `src/private/mx/impl/IdFunctions.h` with the two helpers (`getId`, `setId`) the readers and writers call.
- `ApiCommon.h` explains what ids are for, and that mx repairs a malformed id but does not check uniqueness (see #397).
- A carried-forward (implicit) time signature no longer keeps the previous measure's id; only the measure that states the `<time>` element carries it.
- `SoundData::isSpecified` now counts an id, so a `<sound>` that has nothing but an id is not dropped.

Left out, because placing the id would need a design decision:

- Grouping elements `mx::api` does not model as their own type: direction-type, notations, articulations, ornaments, technical, beam.
- Marks folded into `MarkData`: dynamics, fermata, accidental-mark. These would need new `MarkDataChoice` alternatives.
- harmony and frame: one `DirectionData` holds several `ChordData` that serialize into a single `<harmony>`, so it is not clear which chord owns the id.
- credit: both `<credit>` and `<credit-words>` have an id and `PageTextData` models the pair.
- Elements `mx::api` does not model at all: print, measure-style, grouping, for-part, glissando, slide (see #389).

## Testing

- [x] New `src/private/mxtest/api/IdAttributeApiTest.cpp`: 7 cases, 53 assertions, covering every new field through `roundTrip`, plus one case reading ids straight off source XML and one pinning the implicit-time-signature rule
- [x] `make test-all` green: core roundtrip (839 cases), core unit (44 cases), api/impl suite (6018 assertions in 542 cases), api roundtrip (366 pinned, 0 failed)
- [x] Discovery went from 366 to 368 passing files; `synthetic/bracket.3.1.xml` and `synthetic/dashes.3.1.xml` newly pass because of the bracket and dashes ids, and are pinned in `roundtrip-baseline.txt`. No file regressed.
- [x] `make fmt-check` passes

## References

- Progresses #399
- Related to #397 (id uniqueness is not constrained by `mx::core`)
- Related to #389 (glissando, slide and wavy-line are not modeled, so their ids are not exposed here)
- Requested in discussion #349, which also led to #393

### DIFF
--- a/src/include/mx/api/SpannerData.h
+++ b/src/include/mx/api/SpannerData.h
@@ -5,6 +5,7 @@
 #pragma once
 
 #include "mx/api/ApiCommon.h"
+#include "mx/api/Id.h"
 #include "mx/api/LineData.h"
 #include "mx/api/PositionData.h"
 #include "mx/api/PrintData.h"
@@ -27,8 +28,8 @@ struct SpannerStart
     LineData lineData;
 
     // The id attribute of the element this spanner end is written as -- <bracket>,
-    // <dashes>, or <octave-shift> (see ApiCommon.h).
-    std::optional<std::string> id;
+    // <dashes>, or <octave-shift> (see Id.h).
+    std::optional<Id> id;
 
     SpannerStart() : number{}, tickTimePosition{0}, positionData{}, printData{}, lineData{}, id{}
     {
@@ -43,8 +44,8 @@ struct SpannerStop
     LineData lineData;
 
     // The id attribute of the element this spanner end is written as -- <bracket>,
-    // <dashes>, or <octave-shift> (see ApiCommon.h).
-    std::optional<std::string> id;
+    // <dashes>, or <octave-shift> (see Id.h).
+    std::optional<Id> id;
 
     SpannerStop() : number{}, tickTimePosition{0}, positionData{}, lineData{}, id{}
     {
--- a/src/include/mx/api/SpannerData.h
+++ b/src/include/mx/api/SpannerData.h
@@ -10,6 +10,7 @@
 #include "mx/api/PrintData.h"
 #include "mx/api/SpannerNumber.h"
 
+#include <optional>
 #include <string>
 
 namespace mx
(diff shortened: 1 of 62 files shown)


src/include/mx/api/SpannerData.h
```diff
@@ -25,7 +26,11 @@ struct SpannerStart
     PrintData printData;
     LineData lineData;
 
-    SpannerStart() : number{}, tickTimePosition{0}, positionData{}, printData{}, lineData{}
+    // The id attribute of the element this spanner end is written as -- <bracket>,
```
webern, line 30: End? This is a spammer start.

Issue #412, opened 2026-08-23
model hammer-on and pull-off as paired spanners in `mx::api`


webern:
## What

`<hammer-on>` and `<pull-off>` are paired, note-attached spanners inside `<technical>`, carrying a
required `type` (start/stop), an optional `number`, and text content for the printed label:

```xml
<technical><hammer-on type="start" number="1">H</hammer-on></technical>
```

`mx::api` has no representation for either. `MarkType` has both members commented out
(`src/include/mx/api/MarkData.h:142-143`), as do the `Converter::technicalMarkMap` entries
(`src/private/mx/impl/Converter.cpp:330-332`). They are dropped on read and cannot be authored.
`mx::core` models them fully as `core::HammerOnPullOff` behind
`core::TechnicalChoice::Kind::{hammerOn, pullOff}`, so the data is available and thrown away in
`mx::impl`.

## Requirements

1. Read and write `start`/`stop` attached to the notes that carry them.
2. Honor `number` via the existing `SpannerNumber` mechanism.
3. Preserve the element text (the `H` / `P` label).
4. Pin a corpus file in `src/private/mxtest/api/roundtrip-baseline.txt`, e.g.
   `ksuite/k004a_Technical.xml`. Both are `support="none"` in `data/api.features.xml`, and none of
   the five corpus files using them are defended today.

Follow whatever shape #389 settles on for the other note-attached paired spanners. `tap` (the third
member of audit section 1.2) is not in scope; it carries text only and is not a spanner.

Blocks `u_span` eligibility in #393. Related: #389, #185 (closed; deferred these).

PR #414, opened 2026-08-23
feat: model glissando, slide, and wavy-line spanners in mx::api

### LINKED ISSUE #389: api: model note-attached paired spanners (glissando, slide, wavy-line)

webern:
## Human Summary

Adds glissando, slide and wavy-line to mx::api. These were relatively similar to established patterns and thus straightforward.

## Summary

`glissando` and `slide` were dropped entirely on read and could not be authored: `NoteFunctions::parseNotations` had an empty `break` for glissando and a commented-out body for slide, and the write side had no `CurveType`-style path for either. `wavy-line` read and wrote as an ordinary `MarkType::wavyLine` mark carrying only position/print data; `core::WavyLine::type()` is a required `StartStopContinue` the writer never set, so every wavy line came out `type="start"` regardless of source, and a trill extension read as a start/stop pair was written back as two starts that never close.

This adds all three as proper note-attached spanners:

- `GlissandoStart`/`GlissandoStop` (new `GlissandoData.h`): one family for `<glissando>` and `<slide>`, discriminated by a `GlissandoType` field, following the same pattern `CurveType` already uses to share one shape between `<slur>` and `<tied>`. Carries pairing, the printed text (the element value, e.g. "gliss."), line appearance (`line-type`/`dash-length`/`space-length`), position, font, and color, plus slide-only trill-sound playback (`accelerate`/`beats`/`first-beat`/`last-beat`).
- `WavyLineStart`/`WavyLineContinue`/`WavyLineStop` (new `WavyLineData.h`): `<wavy-line>` is a start/continue/stop spanner living inside a note's `<ornaments>`, not `<notations>` directly. Carries placement, color, an optional SMuFL glyph override, and its own trill-sound playback (`start-note`/`trill-step`/`two-note-turn`/`accelerate`/`beats`/`second-beat`/`last-beat`).
- Both go through the writer-side `SpannerNumber` identity mechanism from #297/#350: `SpannerNumberResolver` gained `glissando`, `slide`, and `wavyLine` number pools, independent from each other and from the existing slur/tie/wedge/octave-shift/bracket/dashes pools, and independent of each other even when `<glissando>` and `<slide>` overlap in the stream.
- `mx::impl`: new `GlissandoFunctions.h` (templated over `core::Glissando`/`core::Slide`, mirroring `CurveFunctions.h`) and `WavyLineFunctions.{h,cpp}`, wired through `NoteFunctions`, `OrnamentsFunctions`, and `NotationsWriter`. Stops are always emitted before starts on a shared note, per the #139 rule.
- Removed `MarkType::wavyLine` and the dead `SlideFunctions` stub (already commented out, referencing a `SpannerType` enum deleted years ago). This is a breaking change: `MarkType::wavyLine`'s mark-based model could not express pairing at all, so it could never be more than a fallback.

## Testing

- [x] New `GlissandoApiTest.cpp` and `WavyLineApiTest.cpp`: 13 cases covering round-trip pairing, printed text, line appearance, slide/wavy-line playback attributes, container placement (`<notations>` vs `<ornaments>`), stop-before-start ordering on a chained note, independent identity-number pools, and `fromXml` reads of the synthetic fixtures
- [x] `make test-all` green: core roundtrip, core unit, api/impl suite (6159 assertions in 560 test cases), api roundtrip (379 pinned, 0 failed)
- [x] Discovery went from 366 to 379 passing files; 13 newly-passing files are attributable to this change and pinned in `roundtrip-baseline.txt`, including the three files the issue names (`lysuite/ly33h_Spanners_Glissando.xml`, `ksuite/k005a_Spanners_Slides.xml`, and progress on `ksuite/k003a_Ornaments.xml`, whose wavy-line mismatch is now fixed though it still fails on an unrelated pre-existing tremolo attribute bug). No file regressed.
- [x] `make fmt-check` passes
- [x] A unity build (`CMAKE_UNITY_BUILD=ON`, batch size 0) compiles clean

## References

- Closes #389
- Uses the spanner identity/numbering mechanism from #297 and #350
- Applies the stop-before-start ordering rule from #139
- Related: #412 (hammer-on/pull-off), which is explicitly deferred to follow whatever shape this PR settles on


src/include/mx/api/WavyLineData.h
```diff
@@ -0,0 +1,135 @@
...
+    unspecified,
+    whole,
+    half,
+    none
+};
+
+// A wavy line indicates a trill extension or a vibrato line. Unlike glissando and slide it can
+// have a continue point (for a line that crosses a system or page break), so it is a
+// start/continue/stop spanner rather than a start/stop one. It lives inside a note's <ornaments>,
+// not directly under <notations>.
```
webern:
```suggestion
// start/continue/stop spanner rather than a start/stop one. It lives inside a note's <ornaments>.
```

src/private/mxtest/api/roundtrip-baseline.txt
```diff
@@ -644,3 +644,22 @@ lysuite/ly45f_Repeats_InvalidEndings.xml
 # <dashes> reaches the api on SpannerStart/SpannerStop, so it is no longer dropped.
 synthetic/bracket.3.1.xml
 synthetic/dashes.3.1.xml
+
+# Unblocked by #389 (glissando/slide/wavy-line spanners). Glissando and slide were
+# dropped entirely on read and cannot be authored; wavy-line read and wrote as an
```
webern:
```suggestion
# dropped entirely on read and could not be authored; wavy-line read and wrote as an
```

PR #415, opened 2026-08-23
feat: derive octave-shift stop size from its start

### LINKED ISSUE #372: Make octave-shift stop size automatic in mx::api
## Problem

`OttavaStart::ottavaType` determines the complete octave shift, including its size:

- 8va / 8vb → `size="8"`
- 15ma / 15mb → `size="15"`
- 22ma / 22mb → `size="22"`

However, `OttavaStop` separately exposes `std::optional<int> size`. Authors must repeat the start's size on the stop when they want it emitted. This creates two fields for one musical fact, and the values can contradict each other.

Omitting the stop size is valid MusicXML, but some importers, including MuseScore, expect it to be present. Authoring interoperable output should not depend on manually copying the start size to the stop.

Using `OttavaType` directly on `OttavaStop` would not be ideal. A MusicXML stop has no up/down direction of its own, so the reader cannot distinguish 8va from 8vb using the stop element alone.

## Proposed Direction

Derive an octave-shift stop's size automatically from its matching `OttavaStart`.

The writer already performs a part-wide spanner pass through `SpannerNumberResolver`. That pass could be extended, or generalized, to associate each ottava stop with its start and make the resolved size available to `DirectionWriter::emitOttavaStop`.

The intended authoring behavior would be:

1. Set `OttavaStart::ottavaType`.
2. Add the corresponding `OttavaStop`.
3. The writer emits the appropriate stop size automatically.

`OttavaStop::size` could then be removed or replaced with a fidelity policy such as:

```cpp
Bool writeSize = Bool::unspecified;
```
Possible semantics:

- unspecified: automatically write the size derived from the start.
- yes: explicitly write the derived size.
- no: omit the size attribute.

The reader could set yes or no according to whether the source stop explicitly contained size, preserving that spelling when required. Authored data would leave the field unspecified and get interoperable output automatically.

## Design Considerations

- This would be a breaking public API change if OttavaStop::size is removed.
- Explicit, identity-based, and unspecified spanner numbers all need defined pairing behavior.
- Pairing must work across measures and with overlapping octave shifts.
- A dangling stop must have a harmless fallback and must not throw. Writing the MusicXML default size of 8 is one possible fallback.
- If a source stop's size contradicts its matching start, the start should probably be authoritative and the output normalized.
- MusicXML permits any positive integer for size, although mx::api models the conventional 8, 15, and 22 values. Nonstandard sizes therefore need a documented normalization policy.

A larger alternative would be to model an ottava as one span containing both endpoints and one OttavaType, but that is outside the scope of this change. (I have proposed this for other spanners as well, but it is a big refactor.)

  ## Acceptance Criteria

  - [ ] An authored ottava requires its type/size to be specified only on OttavaStart
  - [ ] Matching stops automatically emit the start's size
  - [ ] 8va, 8vb, 15ma, 15mb, 22ma, and 22mb are covered
  - [ ] Pairing works across measures and for overlapping numbered or identity spanners
  - [ ] Source presence or absence of the stop size has a documented fidelity policy
  - [ ] Contradictory and dangling endpoints have defined, non-throwing behavior
  - [ ] Existing pinned API round trips remain passing
  - [ ] Focused reader/writer and full API round-trip tests cover the behavior

  ## Acceptance Criteria

- [ ] An authored ottava requires its type/size to be specified only on OttavaStart
- [ ] Matching stops automatically emit the start's size
- [ ] 8va, 8vb, 15ma, 15mb, 22ma, and 22mb are covered
- [ ] Pairing works across measures and for overlapping numbered or identity spanners
- [ ] Source presence or absence of the stop size has a documented fidelity policy
- [ ] Contradictory and dangling endpoints have defined, non-throwing behavior
- [ ] Existing pinned API round trips remain passing
- [ ] Focused reader/writer and full API round-trip tests cover the behavior

webern:
## Human Summary

Automatically writes the ottava stop size as required by MuseScore. Retains round-trip fidelity through a suppression flag.

## Summary

An ottava stated its size twice. `OttavaStart::ottavaType` implies it, and `OttavaStop::size` repeated it, so the two could contradict each other and an author who wanted the interoperable spelling (MuseScore and others expect a size on the stop) had to copy the start's value by hand.

`OttavaStop::size` is replaced by a fidelity knob, `Bool writeSize`. The size itself now comes from the start the stop closes.

- The writer's part-wide spanner pass, generalized from `SpannerNumberResolver` to `SpannerResolver`, pairs each ottava stop with its start and records that start's size alongside the resolved spanner numbers. Pairing is by `SpannerNumber` in serialized order (identity endpoints by id, explicit endpoints by level, unspecified endpoints with each other), so it works across measures and keeps overlapping ottavas apart.
- `unspecified` (the authoring default) and `yes` both write the derived size; `no` omits the attribute. The reader sets `yes` or `no` according to whether the source spelled it out, so existing documents round-trip unchanged.
- A stop with no matching start writes MusicXML's default size of 8 rather than throwing, and a source stop whose size contradicts its start is normalized to the start's value.
- Nonstandard sizes keep their existing normalization (22 exactly, anything else above 8 as 15, everything else 8), now documented at the point of the mapping.
- `ottavaTypeSize` in a new `impl/OttavaFunctions.h` is the single source of the size fact, used by both the start emitter and the stop pairing.

Breaking public api change: `OttavaStop::size` is gone. Authoring an ottava is now start type, stop, done.

Also pins two files in the api round-trip baseline that discovery reported passing. Neither contains an `octave-shift`, so they were unblocked by earlier work and simply had not been pinned.

## Testing

- [x] New `OttavaSizeApiTest` covers all six lines across measures, overlapping numbered and identity spanners, sequential unnumbered lines, the dangling-stop fallback, both `writeSize` settings, and contradiction normalization (`*OttavaSize*`: 116 assertions in 8 test cases)
- [x] `DirectionReader`/`DirectionWriter` unit tests updated, including a new reader case for a stop with no size attribute
- [x] `make api-test` (6171 assertions in 556 test cases)
- [x] `make api-roundtrip` (368 passed, 0 failed of 368 pinned)
- [x] `make core-roundtrip-test` and `make core-unit`
- [x] `make fmt-check`
- [x] Unity build of the `mx` target (`CMAKE_UNITY_BUILD=ON`, batch size 0)

## References

- Closes #372
- Related to #393 (modeling a spanner as one object rather than a start/stop pair, the larger alternative the issue sets aside)

### DIFF
--- a/src/include/mx/api/OttavaData.h
+++ b/src/include/mx/api/OttavaData.h
@@ -50,18 +50,12 @@ class OttavaStop
     SpannerStop spannerStop;
 
     // Most users can ignore this; leave it unspecified. It only controls whether the stop end of an
-    // octave-shift line repeats the line's size (8, 15, or 22). The size is not the stop's own
-    // fact: it belongs to the OttavaStart that opened the line, so an authored ottava states its
-    // octave shift once, on the start, and the writer puts the matching size on the stop.
+    // octave-shift line repeats the line's size (8, 15, or 22). The size is not required on the
+    // stop and carries no new information. Omitting it is legal, but but some importers (MuseScore)
+    // expect it. Therefore `mx::api` defaults to automatically writing the redundant stop size.
     //
-    // unspecified (the default) and yes both write that size; no omits it. Omitting it is legal
-    // MusicXML, but some importers (MuseScore among them) expect it, so the default is the
-    // interoperable spelling. It exists for round-trip fidelity - reading a file sets yes or no
-    // according to whether the source spelled the attribute out.
-    //
-    // The start a stop closes is found by spannerStop.number: an identity, an explicit level, or an
-    // unspecified number pairs with the nearest still-open start of the same kind, within the part
-    // and across measures. A stop that matches no start writes MusicXML's default size of 8.
+    // For the sake of file round-trip fidelity, this will be set to 'no' if a the file being read
+    // does not include the size attribute in the stop element.
     Bool writeSize;
 
     OttavaStop() : spannerStop{}, writeSize{Bool::unspecified}
--- a/src/include/mx/api/OttavaData.h
+++ b/src/include/mx/api/OttavaData.h
@@ -3,6 +3,7 @@
 // Distributed under the MIT License
 
 #pragma once
+#include "mx/api/ApiCommon.h"
 #include "mx/api/SpannerData.h"
 
 namespace mx
@@ -47,13 +48,21 @@ class OttavaStop
   public:
     SpannerStop spannerStop;
 
-    // MusicXML's octave-shift allows a size attribute ("8", "15", or "22") on the stop, not just the
-    // start. Most writers omit it there since it is implied by the corresponding start, but some
-    // importers (e.g. MuseScore) expect it to be present. Absent by default; set it to have the
-    // writer emit it explicitly.
-    std::optional<int> size;
+    // Octave-shift size fidelity knob. A stop's size (8, 15, or 22) is never its own musical fact:
+    // it belongs to the OttavaStart that opened the line, so the writer takes it from there. The
+    // stop is paired with its start by spannerStop.number -- an identity, an explicit level, or an
+    // unspecified number pairs with the nearest still-open start of the same kind, within the part
+    // and across measures.
+    //
+    // unspecified (the default) and yes both write the size the start implies; no omits the
+    // attribute. Omitting it is legal MusicXML, but some importers (e.g. MuseScore) expect it, so
+    // authoring leaves this alone and gets the interoperable spelling. The reader sets yes or no to
+    // record whether the source spelled the attribute out.
+    //
+    // A stop with no matching start writes MusicXML's default size of 8.
+    Bool writeSize;
 
-    OttavaStop() : spannerStop{}, size{std::nullopt}
+    OttavaStop() : spannerStop{}, writeSize{Bool::unspecified}
     {
     }
 };
@@ -67,7 +76,7 @@ MXAPI_NOT_EQUALS_AND_VECTORS(OttavaStart);
 
 MXAPI_EQUALS_BEGIN(OttavaStop)
 MXAPI_EQUALS_MEMBER(spannerStop)
-MXAPI_EQUALS_MEMBER(size)
+MXAPI_EQUALS_MEMBER(writeSize)
 MXAPI_EQUALS_END;
 MXAPI_NOT_EQUALS_AND_VECTORS(OttavaStop);
 } // namespace api
--- a/src/include/mx/api/OttavaData.h
+++ b/src/include/mx/api/OttavaData.h
@@ -31,11 +31,12 @@ class OttavaStart
     SpannerStart spannerStart;
     OttavaType ottavaType;
 
-    // Octave-shift size fidelity knob. The size value (8, 15, or 22) follows from ottavaType, so
-    // 15ma/15mb and 22ma/22mb lines always write their size while an 8va/8vb line omits the
-    // redundant, spec-default size="8". When true, the writer also emits that redundant size="8";
-    // the reader sets it when the source spelled the attribute out. Leave false (the default) when
-    // authoring. It has no effect on a 15ma/15mb or 22ma/22mb line, whose size is always written.
+    // Most users can ignore this; leave it false. It only controls whether an 8va/8vb line writes
+    // the redundant size="8" attribute. The size (8, 15, or 22) follows from ottavaType, and 8 is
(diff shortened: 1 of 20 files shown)


src/private/mx/impl/DirectionReader.cpp
```diff
@@ -720,7 +720,11 @@ void DirectionReader::parseOctaveShift(const core::DirectionType &directionType)
         api::OttavaStop stop;
         stop.spannerStop = impl::getSpannerStop(octaveShift);
         stop.spannerStop.tickTimePosition = myCursor.tickTimePosition;
-        stop.size = octaveShift.size();
+
+        // The stop's size is not kept: it restates the start's octave shift, and a stop whose size
+        // contradicts its start would let the api hold two answers to one question. Only whether
+        // the source spelled the attribute out is recorded, so the same spelling is written back.
```
webern:
```suggestion
        // The api holds size information at the start since specifying it on the stop is redundant.
        // Here we ignore the value and just record its presence or absence for round-trip fidelity.
```

src/private/mx/impl/DirectionReader.cpp
```diff
@@ -740,6 +744,10 @@ void DirectionReader::parseOctaveShift(const core::DirectionType &directionType)
     // to the "down" variants (o8vb/o15mb/o22mb).
     bool isUp = octaveShift.type().tag() == core::UpDownStopContinue::Tag::up;
 
+    // MusicXML allows any positive size, but notation only uses 8, 15, and 22, which is all
+    // OttavaType names. A size outside those three is normalized to the nearest named line it can
+    // stand for -- 22 exactly, anything else above 8 as 15, everything else 8 -- and the line is
+    // written back with that normalized size.
```
webern:
```suggestion
    // MusicXML does not properly constrain the ottava size to valid music notation values. We do
    // so here by interpolating the value into an enum that maps to valid music notation.
```

src/private/mx/impl/DirectionWriter.h
```diff
@@ -23,11 +23,12 @@ namespace impl
 class DirectionWriter
 {
   public:
-    // inNumberResolver supplies the resolved 'number' for each wedge, octave
-    // shift, bracket, and dashes start/stop (see SpannerNumberResolver); it
-    // must outlive this writer.
+    // inSpannerResolver supplies the resolved 'number' for each wedge, octave
+    // shift, bracket, and dashes start/stop, and the size each octave-shift
+    // stop inherits from its start (see SpannerResolver); it must outlive this
+    // writer.
```
webern:
```suggestion
    // inSpannerResolver supplies the resolved 'number' for start/stop pairs such as wedge,
    // octave-shift, bracket, and dashes and stop semantics such as the size attribute of an
    // octave-shift (see SpannerResolver); it outlives this writer.
```

src/private/mx/impl/ScoreWriter.h
```diff
@@ -39,17 +39,18 @@ class ScoreWriter
...
     api::SystemData getSystemData(int measureIndex) const;
 
-    // Spanner 'number' assignments for the ScoreData held by this writer,
-    // computed once at construction (after the score is sorted, so the
-    // resolved object addresses match what the measure/note writers visit).
-    inline const SpannerNumberResolver &getSpannerNumberResolver() const
+    // Spanner 'number' assignments and octave-shift stop sizes for the
+    // ScoreData held by this writer, computed once at construction (after the
+    // score is sorted, so the resolved object addresses match what the
+    // measure/note writers visit).
```
webern:
```suggestion
    // Get the writers shared spanner resolution state object.
```

src/include/mx/api/OttavaData.h
```diff
@@ -47,13 +49,22 @@ class OttavaStop
...
+    // octave shift once, on the start, and the writer puts the matching size on the stop.
+    //
+    // unspecified (the default) and yes both write that size; no omits it. Omitting it is legal
+    // MusicXML, but some importers (MuseScore among them) expect it, so the default is the
+    // interoperable spelling. It exists for round-trip fidelity - reading a file sets yes or no
+    // according to whether the source spelled the attribute out.
+    //
+    // The start a stop closes is found by spannerStop.number: an identity, an explicit level, or an
+    // unspecified number pairs with the nearest still-open start of the same kind, within the part
+    // and across measures. A stop that matches no start writes MusicXML's default size of 8.
```
webern:
```suggestion
    // Most users can ignore this; leave it unspecified. It only controls whether the stop end of an
    // octave-shift line repeats the line's size (8, 15, or 22). The size is not required on the
    // stop and carries no new information. Omitting it is legal, but but some importers (MuseScore)
    // expect it. Therefore `mx::api` defaults to automatically writing the redundant stop size.
    //
    // For the sake of file round-trip fidelity, this will be set to 'no' if a the file being read
    // does not include the size attribute in the stop element.
```

Issue #432, opened 2026-08-27
Consider a different error or warning mechanism.


webern:
Currently `mx` can return a `Result<T>` object during translation, or it can throw. There is no nuance there and there is no way to use the `Result` mechanism from inside the reading and writing mechanisms. A better mechanism would allow an accumulation of errors, warnings and info statements depending on the severity of the issue encountered.

For example, in `SpannerResolver.cpp`, this is not well-aligned with the doctrine, nor would silence be a great option either:

```c++
case api::SpannerNumber::Kind::identity: {
        const auto iter = myResolved.find(inObject);
        if (iter == myResolved.cend())
        {
            MX_THROW("identity spanner was not visited by resolvePart; the writer would emit an "
                     "unnumbered spanner that could collide with another");
        }
        return iter->second;
    }
```

What might be better here is to add an error object to a context (or send it through a lambda in the case of a multithreaded application with, e.g. a live UI), that describes the error and when and where it was encountered. Then translation could continue and the caller could inspect their errors, warnings, and infos and let the application.

Throwing could then be reserved for cases where the translation really cannot proceed because the state is too bad to continue.

Note: the above example might not be the best one since perhaps that *is* a state that is too bad to continue, but the idea stands. Not every little mistake in the score data requires throwing, but it is the only mechanism we have.

PR #435, opened 2026-09-12
feat: replace DocumentManager with RAII MusicXml

### LINKED ISSUE #95: Use RAII for the DocumentManager
The `DocumentManager` is a weird concept. Explain why it exists and how to use it early in the Readme and in the code itself.

For extra credit, return an RAII object instead of a document ID so that leaks won't ever happen.

e.g. `std::unique_ptr<mx::api::Document> createFromFile( "/some/file.xml" )`.

webern:
## Human Summary

This fixes a very old and very bad design, the `DocumentManager` is no more. This should be a little more RAII friendly, though the footguns in C++ are pretty astounding after writing Rust for years.

## Summary

`DocumentManager` was a singleton returning numeric document ids, holding a map of `shared_ptr<Document>`, and requiring the client to call `destroyDocument` by hand. The registry, the ids, the `badDocumentId` error code, and the `getUniqueId` counter all existed only to support that scheme.

They are gone. The public entry point is now a move-only `mx::api::MusicXml` that owns its `core::Document` and frees it at scope exit:

```cpp
auto doc = MusicXml::fromFile(path);            // Result<MusicXml>
auto score = intoScore(std::move(doc).value()); // consumes the doc, frees the tree at return
auto doc = fromScore(score);                    // Result<MusicXml>; can fail with a write refusal
doc.writeToFile(out);                           // Result<void>
auto copy = doc.clone();                        // deep copy
auto& core = doc.getCoreDocument();             // escape hatch: edit the core DOM directly
```

The free functions are the score conversions: `getScore` (const, non-consuming), `intoScore` (by value, consumes and frees the tree at return), and `fromScore` (authors a new document; can fail with the core writer's refusal errors). All entry points preserve the `Result` error boundary: parse/serialize failures return `ioError`, `xmlSyntaxError`, the mirrored core parse codes, or `internalError` for any caught exception. A moved-from `MusicXml` holds a valid, empty document rather than a null pimpl, so every operation on it stays safe. `getCoreDocument` is an escape hatch for when `mx::api` does not do what you need and you want to edit the core DOM directly; it requires including the private `mx::core` headers and is not recommended (open an issue first).

`PartWriter`'s synthesized `<score-instrument>` ids used to come from `DocumentManager::getUniqueId`; they are now a process-wide `std::atomic<int>` counter inside `PartWriter`, seeded at the same high value so they still avoid collisions with ids already present in parsed documents.

### Scope

- New: `src/include/mx/api/MusicXml.h`, `src/private/mx/api/MusicXml.cpp`.
- Removed: `src/include/mx/api/DocumentManager.h`, `src/private/mx/api/DocumentManager.cpp`, `src/private/mxtest/api/DocumentManagerTest.cpp`.
- `Result.h`: `badDocumentId` removed; boundary comments updated to `MusicXml`.
- Tests, examples, and the README ported from the id/`destroyDocument` dance to RAII ownership. `DocumentManagerTest` is replaced by `MusicXmlTest`, which also pins the moved-from behavior.
- Only the api/impl layers change; `mx::core`, the generator, and the roundtrip baseline are untouched.

This is a breaking change to `mx::api`. Client code (`komp`, `denigma`) that called `DocumentManager::getInstance()` must move to `MusicXml` / `fromScore` / `getScore` / `intoScore`.

## Testing

- `MX_RUNNING_IN_DOCKER=1 make api-build` builds mx + mxtest + the three examples + the api-roundtrip binary.
- `MX_RUNNING_IN_DOCKER=1 make api-test`: all tests passed (5495 assertions in 602 test cases). All three api examples ran successfully.
- `make fmt-check` passes.
- `make fmt` / `fmt-check` could not run locally (the lima docker VM mounts `~` read-only, so the container cannot run `clang-format -i` into the source tree); the CI fmt-check will run.

## References

- Closes #95
- Progresses the error-boundary work tracked by #432 (the `MusicXml` boundary preserves the `Result` quarantine `DocumentManager` had).


src/include/mx/api/MusicXml.h
```diff
@@ -0,0 +1,83 @@
...
+using DocumentPtr = std::shared_ptr<Document>;
+} // namespace core
+
+namespace api
+{
+// A MusicXML document that you own. Parse one from a file or a stream, or
+// author one from ScoreData, then read the score back out of it or write it
+// to disk. The document is freed automatically when it goes out of scope;
+// mx keeps no registry of documents and tracks no ids. A document cannot be
+// copied, only moved.
```
webern: Don't overdo it:

> A MusicXML document either parsed or constructed from ScoreData.

src/include/mx/api/MusicXml.h
```diff
@@ -0,0 +1,83 @@
...
+class MusicXml
+{
+  public:
+    // parses a .musicxml file. Errors: ioError (file open/read),
+    // xmlSyntaxError (the bytes are not XML), or the mirrored core parse
+    // errors.
```
webern: Let the Error type documentation explain what the variants are.

> Parses a MusicXML file. Logical errors and caught exceptions are represented by an error result.

src/include/mx/api/MusicXml.h
```diff
@@ -0,0 +1,83 @@
...
+
+    // access to the underlying core document for requirements that ScoreData
+    // does not meet. Prefer the score functions above; the core model is a
+    // much larger interface and it is not frozen the way mx::api is.
+    const core::Document &getCoreDocument() const;
+
+  private:
+    MusicXml(core::DocumentPtr &&coreDocument, bool writeMxVersion);
+    class Impl;
+    std::unique_ptr<Impl> myImpl;
```
webern, line 67: I don't know where to put this, but when the `MusicXml` instance is moved (i.e. destroyed) by passing it to the function which converts it to ScoreData, the pimpl must not be left in a dangerous null state. A default constructed non-null state should always exist here.

src/include/mx/api/MusicXml.h
```diff
@@ -0,0 +1,83 @@
...
+
+    friend Result<ScoreData> getScore(const MusicXml &document);
+    friend Result<MusicXml> fromScore(const ScoreData &score);
+};
+
+// reads the score out of the document. The document stays alive and can be
+// read again or written out. Fails only with internalError.
+Result<ScoreData> getScore(const MusicXml &document);
```
webern: Can this be eliminated? I'm not seeing the use-case, and can't we just offer cloning of the MusicXml object instead (if someone really needs this use case)?

src/include/mx/api/MusicXml.h
```diff
@@ -0,0 +1,83 @@
...
+// is freed when this function returns rather than when your MusicXml binding
+// goes out of scope. Pass the document with std::move, or hand over the
+// Result's value directly. Fails only with internalError.
+Result<ScoreData> intoScore(MusicXml document);
+
+// authors a new document from ScoreData. CAN fail: when the ScoreData
+// describes something the core model will not represent (e.g. more than 8
+// beams), the error is returned rather than silently dropping data.
+Result<MusicXml> fromScore(const ScoreData &score);
+} // namespace api
```
webern, line 88:
```suggestion
Result<MusicXml> fromScore(const ScoreData &score);

} // namespace api
```

src/include/mx/api/Result.h
```diff
@@ -31,7 +31,6 @@ enum class ResultCode
     tooManyElements,
     invalidDocument,
     unsupportedVersion, // mirrored from the core parse boundary
-    badDocumentId,      // handle not in the registry           (api-level)
     internalError,      // caught exception; nothing escapes    (api-level)
```
webern: This is just a request to add the TODO, not to fix it in this PR.

```suggestion
    // TODO: this badly swallows exception information. we need more variants or something like a site and message
    internalError,      // caught exception; nothing escapes    (api-level)
```

src/private/mx/api/MusicXml.cpp
```diff
@@ -13,34 +13,37 @@
...
-    bool writeMxVersion = true;
-};
-
-using DocumentMap = std::map<int, StoredDocument>;
-
-namespace
+// The write side always emits version="4.0" unconditionally: echoing a
+// declared "3.0" (or ScoreData::musicXmlVersion) from a 4.0 model was a
+// fiction. Enforced here at the write boundary on a copy, so the owned
+// document (and the getCoreDocument escape hatch) keeps what was parsed.
```
webern: Reword this. It's hard to understand why this function is here and remove "was a fiction" don't use insulting language like that in comments and don't use comments to describe how something used to be.

Edit: sorry, I think comment wasn't yours and you have only moved it. Please make it better anyway though.

src/private/mx/api/MusicXml.cpp
```diff
@@ -85,61 +88,41 @@ std::string fileExtension(const std::string &filePath)
...
-    {
-        core::ScorePartwise root = copy.asScorePartwise();
-        root.setVersion(std::string{"4.0"});
-        copy.setRoot(core::Document::Root{std::move(root)});
-    }
-    else
+  public:
+    // writeMxVersion governs whether writeTo*() stamps mx's provenance
+    // <software> (see EncodingData::writeMxVersion); it defaults true,
+    // including for parsed documents (whose source never had the stamp).
```
webern:
```suggestion
    // including for parsed documents (whose source did not have the stamp).
```

src/private/mx/api/MusicXml.cpp
```diff
@@ -199,204 +179,176 @@ Result<int> DocumentManager::createFromStream(std::istream &stream)
...
-    try
+    if (myImpl)
     {
-        LOCK_DOCUMENT_MANAGER
+        return *myImpl->document;
+    }
+    // a moved-from document holds nothing; reading it yields an empty core
+    // document rather than a crash
+    static const core::Document emptyDocument{};
+    return emptyDocument;
```
webern: This is good. Can the object itself enforce this invariant? (same as one of my other comments).

src/include/mx/api/MusicXml.h
webern: Done — the class comment is now: 'A MusicXML document, either parsed or constructed from ScoreData.'

src/include/mx/api/MusicXml.h
webern: Done — the error-variant lists are gone from the header; the docs now just say errors are represented by an error result, and the variants live in Result.h.

src/include/mx/api/MusicXml.h
webern: Done — the move operations now enforce this. The Impl holds the core document by value, and both moves leave a valid, empty document behind (move construct allocates the empty replacement, so it is not noexcept; move assignment swaps). All moved-from special cases in the write/read paths are gone; MusicXmlTest pins the behavior.

src/include/mx/api/MusicXml.h
webern: Eliminated. getCoreDocument is gone from the public api and clone() makes a deep copy instead. The core model is not reachable through the public surface at all now; mx's own layers and tests reach it via the new private header mx/api/MusicXmlInternal.h (coreDocumentOf, a friend of MusicXml) — nine test files use it to assert on the core tree.

src/include/mx/api/MusicXml.h
webern: Done — blank line added before the closing namespace.

src/include/mx/api/Result.h
webern: Done — TODO added on internalError.

src/private/mx/api/MusicXml.cpp
webern: Reworded — it now says mx's data model is MusicXML 4.0 so every write states version=4.0, and that the override happens on a copy so the owned document keeps what was parsed.

src/private/mx/api/MusicXml.cpp
webern: Done.

src/private/mx/api/MusicXml.cpp
webern: Yes — same as the pimpl comment: the object now guarantees it, so the moved-from guards in the free functions are gone. See the reply on MusicXml.h:62.

src/include/mx/api/MusicXml.h
webern: Correction to my previous reply: removing it was a mistake, so `getCoreDocument` is back as a public method — the non-const overload allows editing the core DOM, the const overload reads. Documented as an escape hatch per your wording: include the private mx::core headers in your header search path to use it; not recommended, try opening an issue first. The private header indirection (MusicXmlInternal.h) is gone; the tests use the public method again.

webern: Added a lifetime note to the doc comment: the reference is only good while this MusicXml is alive and hasn't been moved away, and specifically calls out intoScore, which frees the tree when it returns — so a reference taken before that call dangles afterward with no compiler diagnostic.

src/include/mx/api/Result.h
```diff
@@ -31,8 +31,9 @@ enum class ResultCode
     tooManyElements,
     invalidDocument,
     unsupportedVersion, // mirrored from the core parse boundary
-    badDocumentId,      // handle not in the registry           (api-level)
-    internalError,      // caught exception; nothing escapes    (api-level)
+    // TODO: this badly swallows exception information. we need more variants
+    // or something like a site and message.
```
webern:
```suggestion
```

src/private/mx/examples/Write.cpp
```diff
@@ -108,30 +108,26 @@ int main(int argc, const char *argv[])
...
     // see issue #150.
     const std::string outputPath = (argc > 1) ? argv[1] : "./example.musicxml";
-    const auto writeResult = mgr.writeToFile(documentID, outputPath);
-
-    // we need to explicitly delete the object held by the manager
-    mgr.destroyDocument(documentID);
+    const auto writeResult = document.writeToFile(outputPath);
 
     return writeResult.ok() ? 0 : 1;
 }
```
webern:
```suggestion
}

```

PR #437, opened 2026-09-16
feat: report non-fatal translation diagnostics

webern:
## Summary

Adds an optional `Diagnostics` collector to the `getScore`, `intoScore`, and `fromScore` translation APIs. Callers can inspect reports after a call or receive them synchronously through a handler while fatal failures remain in `Result`.

The first reports cover two existing recoveries: an out-of-range time-signature staff number being applied to all staves, and an octave-shift stop with no matching start using the MusicXML default size of 8. Diagnostics carry score locations and have a canonical formatter. More than 16 concurrent numbered spanners remains fatal, but now returns a located `tooManyElements` error instead of an unlocated internal error.

## Testing

- [x] Diagnostics tests: 55 assertions in 7 test cases
- [x] `make api-test`: 5580 assertions in 617 test cases
- [x] `make api-roundtrip`: 414 passed, 0 failed
- [x] `make test-all`
- [x] `make wasm-test`
- [x] `make fmt`

## References

- Progresses #432
- Stacked on #436; merges after it


.agents/skills/mx-api-doctrine/SKILL.md
```diff
@@ -42,7 +42,8 @@ Responses to wrong api usage, in order of preference:
 2. Defined fallback: document a harmless result and return it. A wrong-kind choice accessor
    returns a default-constructed copy; the writer drops the half of an encoding that is
    meaningless for the note it is on (a tie on a silent cue note is written as `<tied>`
-   notation only, never as a sound-level `<tie>`). No signal to the caller.
+   notation only, never as a sound-level `<tie>`). Report an important adjustment through an
+   optional `Diagnostics` collector; ignoring diagnostics must leave a safe, usable value.
```
webern:
```suggestion
   optional `Diagnostics` collector.
```

src/include/mx/api/Diagnostics.h
```diff
@@ -0,0 +1,64 @@
...
+    info,    // noteworthy normalization without meaningful loss
+    warning, // adjusted data; the result remains usable
+    error    // omitted data; the operation still produced a value
+};
+
+// The recoverable decision reported by a diagnostic.
+enum class DiagnosticCode
+{
+    valueAdjusted,
+    unmatchedSpanner
```
webern: document

src/include/mx/api/MusicXml.h
webern, line 25: Add usage examples here. One with the provided, post-translation diagnostics collector, and one that passes a callback.

src/private/mx/api/LocationFormatting.cpp
```diff
@@ -0,0 +1,49 @@
...
+// Copyright (c) by Matthew James Briggs
+// Distributed under the MIT License
+
+#include "mx/api/LocationFormatting.h"
+
+namespace mx
+{
+namespace api
+{
+void appendFormattedLocationAndMessage(std::string &text, const Location &location, const std::string &message)
```
webern: The distinction between what `text` is and what `message` is seems blurry to me.

src/private/mx/impl/DiagnosticsContext.h
```diff
@@ -0,0 +1,50 @@
...
+        location.partIndex = cursor.partIndex;
+        location.measureIndex = cursor.measureIndex;
+        location.staffIndex = cursor.staffIndex;
+        location.voiceIndex = cursor.voiceIndex;
+        location.tickTimePosition = cursor.tickTimePosition;
+        report(severity, code, std::move(location), std::move(message));
+    }
+
+  private:
+    api::Diagnostics *myDiagnostics = nullptr;
```
webern: Let's use safe, modern pointer objects.

src/private/mx/impl/ScoreReader.cpp
```diff
@@ -268,7 +268,7 @@ api::ScoreData ScoreReader::getScoreData() const
     for (const auto &reconciledPart : partMap)
     {
         PartReader reader{*reconciledPart.first, *reconciledPart.second, myOutScoreData.ticksPerQuarter,
-                          myScorePartwise, divisionsValue};
+                          myScorePartwise,       divisionsValue,         myDiagnostics};
```
webern: Is this fmt'ed? It looks weird.

src/private/mx/impl/SpannerResolver.cpp
```diff
@@ -368,8 +397,10 @@ static void spannerNumberAssignClass(const std::vector<SpannerNumberEvent> &inEv
         }
         if (chosen == 0)
         {
-            MX_THROW("more than 16 spanners of one type are open at the same point in the serialized "
```
webern, line 371: Can we update the `MX_THROW` macro to accommodate the new shape instead of not using it?

src/private/mx/impl/MeasureReader.cpp
```diff
@@ -852,7 +856,7 @@ void MeasureReader::parseFiguredBass(const core::FiguredBass &inMxFiguredBass, c
 
     direction.figuredBasses.emplace_back(std::move(figuredBass));
 
-    const auto staffIndex = getFiguredBassStaffIndex(myCurrentCursor, myOutMeasureData, nextNotePtr);
+    const auto staffIndex = measureReaderFiguredBassStaffIndex(myCurrentCursor, myOutMeasureData, nextNotePtr);
```
webern: What's going on with these renames and the ad-hoc namespacing prefix `measureReader`? Is there something more idiomatic that could be done here? Was a name collision introduced? Can this be cleaner?

.agents/skills/mx-api-doctrine/SKILL.md
webern: Fixed in `21f715cc`: shortened the doctrine entry to the suggested wording.

src/include/mx/api/Diagnostics.h
webern: Fixed in `21f715cc`: documented both diagnostic codes, the diagnostic fields, the callback type, and the collector methods.

src/include/mx/api/MusicXml.h
webern: Fixed in `21f715cc`: added post-translation collection and callback usage examples beside the diagnostics overload.

src/private/mx/api/LocationFormatting.cpp
webern: Fixed in `21f715cc`: replaced the mutating helper with `formatLocationAndMessage`, which returns the location/message suffix; callers now build their own complete result string.

src/private/mx/impl/DiagnosticsContext.h
webern: Fixed in `21f715cc`: `DiagnosticsContext` now carries a shared `reference_wrapper` instead of a raw pointer.

src/private/mx/impl/ScoreReader.cpp
webern: Fixed in `21f715cc`: introduced named local arguments so clang-format leaves the constructor call clean and unpadded. `make fmt-check` passes.

src/private/mx/impl/SpannerResolver.cpp
webern: Fixed in `21f715cc`: added `MX_THROW_AS` for typed exceptions and used it for the located `WriteRefusal`.

src/private/mx/impl/MeasureReader.cpp
webern: Fixed in `21f715cc`: the collision came from the strict unity build exposing generated `parseFigure`; the former anonymous helpers are now private static members of `MeasureReader` and `PartReader`, without ad-hoc prefixes. The strict unity build passes.

PR #460, opened 2026-09-19
feat: model pedal number so overlapping pedal lines keep their identity

### LINKED ISSUE #411: expose pedal/@number in `mx::api::PedalLineData`
`mx::api::PedalLineData` carries only `kind`, `tickTimePosition` and `positionData`. MusicXML's
`<pedal>` also has a `number` attribute (`number-level`), which is what distinguishes concurrent
pedal lines on one staff -- a damper line and a sostenuto line overlapping, for example.

Today the number is dropped in both directions:

- `DirectionReader` reads `pedal.type()`, `pedal.line()` and the position attributes, but never
  looks at `pedal.number()` (`src/private/mx/impl/DirectionReader.cpp:676`).
- `DirectionWriter::emitPedal` sets `type` and `line` and never sets `number`
  (`src/private/mx/impl/DirectionWriter.cpp:322`).

So a score with two overlapping pedal lines round-trips into a single ambiguous run of events, and
an author using `mx::api` has no way to say which `stop` closes which `start`.

## proposed

Give `PedalLineData` a `SpannerNumber number` field -- the same identity type wedges and curves
already use -- and wire it through the reader and through `NumberResolver` in the writer the way
`emitWedgeStart` / `emitWedgeStop` do.

## deliberately not part of this

`<pedal>` has no `line-type`, `dash-length`, `space-length` or `end-length` attributes; it is not in
MusicXML's line-type or dashed-formatting attribute groups (see the generated `core::Pedal` in
`src/private/mx/core/generated/Pedal.h`). There is therefore no `LineData` for `PedalLineData` to
carry, and a consumer that wants pedal line styling has to pick it itself. Recording this here
because a consumer coming from an older fork that modelled pedals as generic `SpannerStart` /
`SpannerStop` will have had a `lineData` member and may expect one back; that member could never
have round-tripped.

`<pedal>` also has `sign` and `abbreviated`. mx currently uses them only to choose between the line
form (`PedalLineData`) and the sign form (`MarkType::pedal` / `MarkType::damp`). Round-tripping
`abbreviated` -- the short `P` / `S` signs versus the full `Ped` / `Sost` -- is a separate gap and
should be a separate issue if it is wanted.

## related

- #393 -- spanner tracking issue. The `number` added here is the same spanner identity that design
  is concerned with.

webern:
## Human Summary

The ability to specify the line level for pedal lines was missing. Added here.

## Summary

`mx::api::PedalLineData` now carries a `SpannerNumber number`, and both translation directions
honour MusicXML's `<pedal number="...">` attribute. The field is the same identity type wedges,
curves and wavy lines already use: leave it unspecified for a lone pedal line, give an explicit
level to write that level verbatim, or give the events of one line a shared identity label and the
writer assigns a level from serialization order.

The reader takes the number off the element in `DirectionReader::parsePedal`. The writer emits it
from `DirectionWriter::emitPedal` through `SpannerResolver`, which now gives pedal lines their own
pool of numbers 1..16 like every other spanner family. A pedal line opens with `start`, `sostenuto`
or `resume` and closes with `stop` or `discontinue`; `change` and `continueLine` happen while the
line stays open.

Nothing changes for an ordinary score: a single pedal line with no number writes no number
attribute, and every other spanner family keeps its own pool and assignments.

Deliberately left out, as separate gaps: `pedal/@sign` and `pedal/@abbreviated` (the sign form is
modelled through `MarkType::pedal` / `MarkType::damp`), the font and color half of
`print-style-align` on `<pedal>`, and numbers on the pedal sign form.

## Testing

- [x] New reader test: `<pedal type="start" line="yes" number="2">` reads as `PedalLineKind::start` with explicit level 2; a pedal with no number reads as unspecified
- [x] New resolver/writer tests: two overlapping identity pedal lines get distinct numbers and both survive a round trip; an explicit pedal level is written verbatim and round-trips; a pedal number and a wedge number come from separate pools
- [x] New round-trip tests: an explicit pedal number comes back; a pedal line without a number writes no `number` attribute
- [x] Reverting just the reader/writer/resolver sites fails 5 of the new test cases (10 assertions), so they are real regressions
- [x] `make api-test`: all pass (5909 assertions in 664 test cases)
- [x] `make api-roundtrip`: 414 passed, 0 failed (of 414 pinned)
- [x] `make api-roundtrip-discover`: 414 PASS, 426 FAIL, 0 LOADFAIL/GETDATAFAIL/CREATEFAIL -- no file unlocked, so no baseline pin. `synthetic/pedal.3.1.xml` still fails because its pedal also carries `abbreviated`, font and color attributes that `mx::api` does not model
- [x] `make test-all`, `make fmt-check`

## References

- Closes #411
- Related to #455
- Related to #393


src/include/mx/api/PedalLineData.h
```diff
@@ -54,14 +55,20 @@ struct PedalLineData
     int tickTimePosition;
     PositionData positionData;
 
+    // Which line this event belongs to when two pedals are held down at once; MusicXML calls it
+    // the number. A sostenuto line pedaled under a damper line needs a different number from it,
+    // so the player can tell which lift ends which line. Give every event of one line the same
+    // number, or leave it unspecified when only one line is open.
```
webern: Can it also be left unspecified when we want the SpannerResolver to handle it for us?

And should it be declared like this?

```
SpannerNumber number = UNSPECIFIED;
```

src/include/mx/api/PedalLineData.h
```diff
@@ -54,14 +55,24 @@ struct PedalLineData
...
     PositionData positionData;
 
+    // Which line this event belongs to when two pedal lines are held down at once; MusicXML calls
+    // it the number. It can be left unspecified, and usually should be: a lone line, and any run
+    // of lines that do not overlap, needs no number. A number is only needed to tell two lines
+    // apart when they overlap -- when a line opens before the one before it has closed, as a
+    // sostenuto line under a damper line does -- because otherwise a lift cannot be matched to
+    // the right downstroke. Give every event of one line the same SpannerNumber (see
+    // SpannerNumber.h): an explicit level is written verbatim, or an identity label lets the
+    // writer assign the level.
```
webern:
```suggestion
    // the right downstroke. In non-ambiguous cases, leaving this unset is fine.
```

Issue #434, opened 2026-08-29
Consider a tuplet authoring helper.


webern:
Highly related to both #428 and #393.

A tuplet isn't a typical spanner in the sense that it modifies the duration of the notes it contains. It would be interesting to consider making Tuplets easy to author by giving them the single start/stop treatment as we contemplate with spanners.

However with tuplets there are some differences. A tuplet stops when it stops, by definition it lasts for a certain amount of musical time.

One thing that will come up with nested tuplets as well: truly nested tuplets don't seem like such a big problem, but in musicxml I believe you could have a "nested" tuplet where the "first" tuplet ends before the "nested" tuplet ends. I imagine someone like Ferneyhough has done this even though I've never seen it.

Anyway the idea is simple on its own, a tuplet object that either contains its notes or where the start conveys all the information such that the user doesn't have to modify each tuplet's note and enter the stop sentinel.

> AI notes from the #428 session:

## AI Notes

What makes tuplets hard to author is not the start/stop sentinels — it is that MusicXML spreads one musical fact across every member note. The author means "3 in the time of 2 eighths, starting here"; the encoding wants a cumulative `<time-modification>` on each note (nested tuplets multiply: 3:2 inside 3:2 is 9:4), `<normal-type>`/`<normal-dot>` only where the tuplet's unit differs from a note's own `<type>`, and paired `<tuplet>` notations with hand-allocated numbers. #428 settled that the writer transcribes these fields rather than inferring them, so the fussy knowledge needs an opt-in home: a helper that writes ordinary fields the author could have set by hand. The writer stays honest, fidelity is untouched, and the automatic experience the old inference gestured at gets delivered whole instead of half.

On shape: containment fights the model twice. Notes are already contained — voice vectors ordered by tick — so a tuplet object owning notes would be a second owner competing with the voice. And the Ferneyhough case rules it out: overlapping tuplets are not a tree, while `<tuplet>` number-pairing is flat and permits them. The start-conveys-all form survives both. The observation that a tuplet lasts a definite musical time is the key: `TupletStart` already carries actual/normal number, duration name, and dots, so the extent is derivable — normalNumber × the normal unit of musical time from the start. Both the stop sentinel and every member's ratio are computable from the start alone.

Two altitudes, not mutually exclusive:

[...]

Issue #429, opened 2026-08-24
NotationsWriter emits all spanner stops before all starts, breaking single-note spanners


rpatters1:
[**NOTE**: This is a more serious issue than just tuplets on single notes. See the glissando examples below.]

`NotationsWriter::writeNotations` writes every entry of `NoteData::noteAttachmentData::tupletStops` before every entry of `tupletStarts`:

```cpp
for (const auto &tupletStop : myNoteData.noteAttachmentData.tupletStops)  { ... }   // all stops
for (const auto &tupletStart : myNoteData.noteAttachmentData.tupletStarts) { ... }  // then all starts
```

That convention is right when a note closes one tuplet and opens another. It is wrong for a tuplet contained in a single note, where both ends belong to the same note and MusicXML wants `start` and then `stop`.

**Symptom.** A Finale document with a nested tuplet whose inner tuplet covers exactly one note (the last note of the outer tuplet) produces, on that note:

```xml
<tuplet type="stop" number="1"/>
<tuplet type="stop" number="2"/>
<tuplet type="start" number="2"/>
```

Number 2 closes before it opens. A reader encounters a stop for a tuplet that was never open, then a start that is never closed. The correct sequence is:

```xml
<tuplet type="start" number="2"/>
<tuplet type="stop" number="2"/>
<tuplet type="stop" number="1"/>
```

The numbers themselves are correct and pair correctly; only the order is wrong.

**Why the author cannot work around it.** `tupletStarts` and `tupletStops` are separate vectors, so there is no way to express "this start precedes that stop on the same note". Any ordering the author chooses within each vector is discarded by the two-loop structure.

**Suggested shape**

Order a note's tuplet notations so that a start whose matching stop is on the same note is emitted before that stop. Since `numberLevel` already pairs the two ends, that can be decided locally: when a `TupletStart` and a `TupletStop` on one note share a `numberLevel`, write the start first, and otherwise keep the existing stops-before-starts order for the tuplets that merely pass through.

**Note**

This is the ordering half only. `<normal-type>` for the same nested-tuplet territory is #428, and the two are independent. Tuplets also still carry a raw `numberLevel` int rather than an `api::SpannerNumber`, so they get none of the writer-side number assignment from #320; that is a third, separate matter and is not the cause of this bug.

rpatters1:
Here is another automatically submitted by Claude. I think I have got it to where it will prompt me before submitting any more. I thought you might find some sample fixtures helpful for this case.

[...]

Issue #428, opened 2026-08-24
NoteWriter infers <normal-type> from sibling notes instead of using DurationData::timeModificationNormalType


rpatters1:
`mx::impl::NoteWriter` does not use `api::DurationData::timeModificationNormalType`. Instead it infers `<normal-type>` by scanning sibling notes for exactly one `TupletStart` and exactly one `TupletStop`. Two consequences follow, both measurable against Finale's own export of the same file.

**Over-emission on ordinary tuplets.** MusicXML reads an absent `<normal-type>` as the note's own `<type>`, so it only needs writing when the tuplet's normal note value differs. An author that sets the field only in that case still gets one on every note of every tuplet. For one Finale test document:

| | `<normal-type>` count |
|---|---|
| Requested through `DurationData` | 11 |
| Written to the file | 253 |
| Of those, merely repeating the note's own `<type>` | 242 |
| Finale's own export of the same music | 4 |

Redundant rather than incorrect, but it makes a diff against a reference export unreadable, and it overrides a deliberate choice the API already lets the author express.

**Silent omission on nested tuplets.** When a note opens two tuplets the sibling search matches nothing, so no `<normal-type>` is written at all, regardless of what the API asked for. For a 3:2 eighth tuplet inside a 3:2 quarter tuplet the omission happens to be right, because the cumulative `9:4` is expressed in the eighths the notes already are. That is luck rather than design: nothing in the writer establishes it, and a nested tuplet whose normal value does differ from the note type has no way to say so.

`DurationData` already models the whole cumulative time modification:

```cpp
timeModificationActualNotes
timeModificationNormalNotes
timeModificationNormalType
timeModificationNormalTypeDots
```

Since `<time-modification>` is cumulative across nesting while `<tuplet>` notations are per-tuplet and visual, the two should not be derived from each other.

**Suggested shape**

- Write `<normal-type>` from `DurationData::timeModificationNormalType` (and its dots) when set, and omit it when unspecified. Treat the field as the author's statement rather than a hint.
- Pair `TupletStart` / `TupletStop` by `numberLevel` instead of assuming one of each per note, so nested tuplets are handled rather than skipped.

**Scope**

This issue is `<normal-type>` only. Two neighbouring tuplet matters are separate and neither is caused by this one:

- #429, a notation ordering bug: `NotationsWriter` writes all tuplet stops before all tuplet starts, which inverts a tuplet contained in a single note.
[...]

PR #463, opened 2026-09-20
feat: tuplet numbering and portion fidelity

webern:
## Human Summary

Adds support for nested tuplets and for controlling more about how a tuplet is spelled in the MusicXML. I'm not entirely sure if the LLM understand how nested tuplets work because I don't know well enough to check the synthetic file it created; but the unblocked wild corpus files more or less prove that users can do what they need to do now.

## Summary

Two related pieces of tuplet work, plus three attributes that had no home in `mx::api`.

**Numbering.** `TupletStart` and `TupletStop` carry a `SpannerNumber` instead of a raw `int
numberLevel`, and tuplets join the other spanner families in `SpannerResolver`. Tuplets were the
last spanner family whose numbers the caller had to allocate by hand, which is exactly what
nested tuplets need. An author who gives both ends of a tuplet the same identity now gets the
numbers assigned for them: nested tuplets overlap in the serialized stream and are handed 1 and
2, while tuplets that follow one another both get 1. Two unnumbered tuplets open at once are
reported through `Diagnostics`, since a reader defaults both to 1 and cannot tell them apart.

**Portion fidelity.** A `<tuplet>` may spell its ratio out in `<tuplet-actual>` and
`<tuplet-normal>` or leave both out and let the note's `<time-modification>` speak for it. The
reader accepted both shapes and threw the difference away; the writer then put both portions on
every start, turning an unspecified note value into a `<tuplet-type>maxima</tuplet-type>` the
source never had. That was the sole remaining round-trip blocker on seventeen wild corpus files.

The reader now records which shape the file used, and the writer states a portion only where it
says something the note does not already say. That rule is what nested tuplets need: an inner
triplet of eighths inside a triplet of quarters gives its notes a cumulative 9:4, so the inner
tuplet's own 3:2 differs from the note and is written, while the outer tuplet repeats its note's
3:2 and is left out. `TupletStart::writeActual` and `writeNormal` are fidelity fields in the
`ClefData::writeStaffNumber` style: unspecified applies the rule, yes/no force it either way, so
a caller who finds the portions noisy can turn them off. Inside a portion, only the parts that
were stated are written.

**New attributes.** `show-type` (two `Bool`s, modeled like `show-number`), `line-shape`
(`TupletLineShape`), and `bracket` on a tuplet stop.

Breaking for callers that set `numberLevel`: `number = SpannerNumber{level}` to write it and
`number.level()` to read it back. Callers that relied on both portions always being written need
`writeActual`/`writeNormal` set to `yes`.

## Testing

- [x] `make api-test` (6078 assertions in 698 test cases), including new `Tuplet` and
      `TupletNumber` suites covering the reader, the writer's rule, the fidelity fields, the
      three attributes, identity and explicit numbering, and both diagnostics
- [x] `make core-roundtrip-test` with the pinned count 840 -> 841
- [x] `make api-roundtrip` 435 of 435 pinned, up from 415
- [x] `make fmt-check`

Twenty files join `roundtrip-baseline.txt`: seventeen wild
(`foundsuite/Invention_5`, `ksuite/k011a_Tuplets`, `lysuite/ly23a`, `ly23b`, `ly23c`, `ly23d`,
`mjbsuite/PreserveTimeModificationNormalType`, the nine `musuite` tuplet and duration-rounding
files, `rpatters1/tuplet_singletons-ref`) and three synthetic, one of which is the new
`data/synthetic/tuplet-nested.3.0.xml` fixture. `ly23d_Tuplets_Nested.xml` is the corpus's one
genuinely nested tuplet.

## References

- Progresses #434. The authoring helper it asks for is still open; this removes the hand-allocated
  numbers that made nested tuplets awkward to write.
- Follows #428: `<normal-type>` is the author's statement, and the same reasoning now applies to
  the tuplet portions.
- Related to #429: a tuplet that starts and stops on one note is still written start-first, now
  paired by `SpannerNumber` rather than by number level.
- Related to #393.


src/include/mx/api/TupletData.h
```diff
@@ -48,40 +59,64 @@ class TupletStart
     // a triplet or the 4 in 5:4 for quintuplet
     Bool showNormalNumber;
 
+    // show the note value of the tuplet's own notes,
```
webern: I can't tell from this how I'm supposed to use it or when I can leave it unspecified. Please use /mx-comments and write this for a human user that does not have your session context. Explain what it is, what it does, when it's required and when it's optional.

src/include/mx/api/TupletData.h
```diff
@@ -48,40 +59,64 @@ class TupletStart
...
+    // for example the eighth note in "3 eighths"
+    Bool showActualType;
+
+    // show the note value of the 'normal' notes,
```
webern: I can't tell from this how I'm supposed to use it or when I can leave it unspecified. Please use /mx-comments and write this for a human user that does not have your session context. Explain what it is, what it does, when it's required and when it's optional.

src/include/mx/api/TupletData.h
```diff
@@ -48,40 +59,64 @@ class TupletStart
...
+
     Bool bracket;
 
+    // Whether the bracket is drawn as a straight line or in the older curved style.
+    TupletLineShape lineShape;
+
+    // Most users can ignore these; leave them unspecified. They control whether the ratio is
+    // spelled out in the file alongside the note's own time modification. unspecified (the
+    // default) applies the right rule automatically: state the ratio when it differs from what
+    // the note already says, which is what an inner tuplet needs, and leave it out otherwise.
```
webern: The introduction of "inner tuplet" without more context is bad. "Inner tuplet" is not a common use case so you have to explain "In the case of nested tuplets..."

